PR #7, opened 2021-02-03
first attempt to add workflow

OnnoEbbens:


dbrakenhoff:
Comments on notebooks:
- 01 contains reference to broken KNMI api, but this is fixed now i believe.
- 07 set extents equal in resampling comparison plots. (+typo in Rectengular)
- 07 some headings could use a bit more explanation, i.e.  how it deals with ties (i.e. when 4 cells are equally close to new cell midpoint)

PR #30, opened 2021-12-06
Update package

OnnoEbbens:
Major update including:
- new way of caching data using decorators in the new caching module
- new plotting methods
- functions to export modeldata to gis in the new gis module
- many smaller bug fixes


nlmod/mdims/mgrid.py
```diff
@@ -269,7 +265,7 @@ def add_idomain_from_bottom_to_dataset(bottom, model_ds, nodata=-999):
 
     idomain = xr.where(bottom.isnull(), -1, 1)
```
dbrakenhoff, line 267: fix comment

nlmod/mdims/resample.py
```diff
@@ -582,16 +511,15 @@ def fillnan_dataarray_structured_grid(xar_in):
     arr_out = values_out.reshape(xar_in.shape)
 
     # bathymetry without nan values
```
dbrakenhoff, line 513: lots of mentions of bathymetry in this generic function

nlmod/mfpackages/constant_head.py
```diff
@@ -0,0 +1,60 @@
...
+@author: oebbe
+"""
+
+import xarray as xr
+import numpy as np
+
+from .. import util, cache
+
+@cache.cache_netcdf
+def get_chd_at_model_edge(model_ds, idomain):
```
dbrakenhoff, line 14: we should also add a method that does this based on a shape(file)?

nlmod/util.py
```diff
@@ -113,265 +113,33 @@ def get_model_ds_empty(model_ds):
...
-def get_cache_netcdf(use_cache, cachedir, cache_name, get_dataset_func,
-                     model_ds=None, check_grid=True,
-                     check_time=True, **get_kwargs):
-    """Create, read or modify cached netcdf files of a model dataset.
-
-    Steps:
 
-    1. Read cached dataset and merge this with the current model_ds if all
-    of the following conditions are satisfied:
+    return model_ds[list(model_ds.coords)]
```
dbrakenhoff, line 117: is this a copy, or is it still linked to the original data?

nlmod/mfpackages/constant_head.py
OnnoEbbens: Yes, that would not be too hard. The function `gdf_to_bool_data_array` from the `mgrid` module can probably help with that.

nlmod/util.py
OnnoEbbens: Good question. I am not 100% sure but I think it is a copy. If I later add data variables to model_ds these are not added to the returned dataset from `get_model_ds_empty`. I also tried modifying the original coördinates of model_ds but could not do this. Therefore I think it is fine to get an empty model dataset using this code. To be sure we can also add `.copy(deep=True)` at the end. This is slower but not extremely slow, see timing below:

image

What do you think?

nlmod/mdims/mgrid.py
OnnoEbbens: fixed

dbrakenhoff (commented):
Looks good! Lots of nice changes I think! I think after tests pass it can be merged, but leaving some things i came across anyway:

Comments on notebooks:
- 01 contains reference to broken KNMI api, but this is fixed now i believe.
- 07 set extents equal in resampling comparison plots. (+typo in Rectengular)
- 07 some headings could use a bit more explanation, i.e.  how it deals with ties (i.e. when 4 cells are equally close to new cell midpoint)

nlmod/mdims/resample.py
OnnoEbbens: fixed

nlmod/mfpackages/constant_head.py
dbrakenhoff: Can be done in as part of a new issue.

nlmod/util.py
dbrakenhoff: Perhaps we can check by checking the memory id for the object, the dataarray and underlying array? `id(model_ds["x"])` etc?

But I'm also fine with the `.copy()` for now just to be sure.

dbrakenhoff:
Also we should mention somewhere this is a breaking update for existing scripts relying on nlmod!

nlmod/util.py
OnnoEbbens: The id's are the same but they differ everytime I call the `id` function which is a bit confusing. I will use `copy()` for now

OnnoEbbens:
Yes we should, I can mention it when I draft a new release

ArtesiaWater (approved):

PR #39, opened 2022-03-17
Update to version 0.2.2

OnnoEbbens:
Minor update including:
- extra options to project vector data on a grid
- function to create a horizontal flow barrier from a line shape
- some bug fixes


nlmod/mdims/mgrid.py
```diff
@@ -790,6 +793,225 @@ def polygon_to_area(modelgrid, polygon, da,
...
+        gdf_agg.set_index(pd.MultiIndex.from_tuples(gdf_cellid.cellid.values),
+                          inplace=True)
+
+    for ind, row in gdf_agg.iterrows():
+        da.values[ind[0], ind[1]] = row[field]
+
+    return da
+
+
+def interpolate_gdf_2_array(gdf, gwf, field='values', method='nearest'):
```
dbrakenhoff, line 861: I would just use the word "to" here :)

dbrakenhoff (approved):

nlmod/mdims/mgrid.py
OnnoEbbens: fair enough :)

OnnoEbbens:
There seems to be an issue with nbconvert (https://github.com/jupyter/nbconvert/issues/1742), that's why the test fail now. The previous test didn't fail and the last commit hardly changed anything so I will merge anyway.

PR #58, opened 2022-05-04
Change the way time is added to a model

rubencalje:
This pull request changes the way time is added to a model:

```
time = pd.date_range('2000', '2000-1-4')
model_ds = nlmod.mdims.set_model_ds_time(model_ds, time=time)
```

```
<xarray.Dataset>
Dimensions:  (time: 4)
Coordinates:
  * time     (time) datetime64[ns] 2000-01-01 2000-01-02 2000-01-03 2000-01-04
Data variables:
    *empty*
Attributes: (12/13)
    time_units:                DAYS
    start_time:                1990-01-01 00:00:00
    nstp:                      1
    tsmult:                    1.0
    steady_start:              1
    steady_state:              0
```

When `steady_start `is True, a steady state stress-period is added before the first value of time with a stress period length of `steady_start_perlen`. The current ways to specify time (`transient_timesteps `or `perlen`) are still supported.

The variable `time `now always contains the time at the end of each stress period (not sure if this was the case before), so heads and other output can be easily added to the model_ds. To minimise double information, `nper `and `perlen `are removed from the attributes of model_ds, and are calculated when needed.


OnnoEbbens (approved):

rubencalje:
> Nice improvements!
> 
> I did add something else when I went through the changes. I moved the attributes with time information ('time_units', 'start_time', 'nstp', 'tsmult', 'steady_start' and 'steady_state') from the model_ds to the time DataArray. So instead of `model_ds.attrs['time_units'] = time_units` you get `model_ds.time.attrs['time_units'] = time_units`.
> 
> I tested locally and it worked. So when the test pass this can be merged.

Good change. Maybe we can then also rename 'time_units' to 'units' and 'start_time' to 'start'?

OnnoEbbens:
Good one, added this in [8e54b1b](https://github.com/ArtesiaWater/nlmod/commit/8e54b1b2332fb96fde6bb2b3c6dc1419c4fae8bb)

PR #85, opened 2022-08-25
Add Modpath

OnnoEbbens:
Add functions to create a modpath model to calculate pathlines. Also add an example notebook and applied it in the Schoonhoven notebook.


OnnoEbbens:
I fixed (ignored) the issues via Codacy but I don't know how to update the Codacy analysis in this pull request other than creating a new useless commit.

rubencalje:
Can you commit the notebooks without output?

A lot of the changes are caused by previous commits by me. It turns out I used a maximum line length of 79 characters, while the defaults line length of the black formatter is 88 characters (which is what you use). I will change my maximum line length to 88 characters, so these changes will not happen in the future anymore.

OnnoEbbens:
I cleared the output of the notebook. Nice you found out about the line length. 

I also struggled a lot with the MemoryFile in the ahn module. I think the problem was that when you create a MemoryFile and read the xarray there is some lazy execution. Xarray does not read the data in the file but stores some kind of reference to that file. When the function returns the Dataarray the MemoryFile is not available anymore, so when you try to access the data in the Dataarray you get an error or an array full of nans. I now force the Dataarray to load the data from the MemoryFile before returning the Dataarray.

PR #88, opened 2022-09-05
Add rotation

rubencalje:
This pull request adds support for rotated grids in nlmod. It is implemented in the folllowing manner:
- angrot, xorigin and yorigin (naming equal to modflow 6) are added to the attributes of the model Dataset.
- angrot is the counter-clockwise rotation angle (in degrees) of the model grid coordinate system relative to a real-world coordinate system (equal to definition in modflow 6)
- when a grid is rotated: x and y (and xv and yv for a vertex grid) are in model-coordinates
- when a grid is rotated: xc and yc are added to the Dataset and represent the cell centers in real-world coordinates (naming equal to rioxarray rotated grids)
- when a grid is rotated: the plot-methods plot the grid in model-coordinates
- when a grid is rotated: geodataframes are automtically transformed to model coordinates before intersecting with the grid

When a grid is not rotated nothing should change.

Grid rotation is now supported in a large part of nlmod, but more work is probably needed.


nlmod/mdims/mgrid.py
```diff
@@ -52,7 +56,7 @@ def xy_to_icell2d(xy, ds):
     return icell2d
 
 
-def modelgrid_from_model_ds(model_ds):
+def modelgrid_from_model_ds(model_ds, rotated=True):
```
OnnoEbbens, line 59: Do we need the `rotated` argument? Is checking for `angrot`==0.0 not enough?

nlmod/mdims/mgrid.py
```diff
@@ -195,15 +223,28 @@ def refine(
...
+        affine_matrix = get_affine_world_to_mod(ds).to_shapely()
+
     if refinement_features is not None:
         for refinement_feature in refinement_features:
             if len(refinement_feature) == 3:
                 # the feature is a file or a list of geometries
                 fname, geom_type, level = refinement_feature
+                if not model_coordinates and ds_has_rotation:
+                    raise (
+                        Exception("Converting files to model coordinates not supported")
```
OnnoEbbens, line 240: NotImplementedError instead of Exception?

nlmod/mdims/resample.py
```diff
@@ -10,6 +10,9 @@
 from scipy.interpolate import griddata
 import rasterio
 from scipy.spatial import cKDTree
+from shapely.geometry import Polygon
+from shapely.affinity import affine_transform
+from affine import Affine
```
OnnoEbbens, line 15: Extra dependency, add to setup.py

nlmod/mdims/resample.py
```diff
@@ -443,11 +493,20 @@ def structured_da_to_ds(da, ds, method="average"):
...
             kwargs["fill_value"] = "extrapolate"
-        da_out = da.interp(x=ds.x, y=ds.y, method=method, kwargs=kwargs)
+        x = ds.x
+        y = ds.y
+        # if has_rotation:
+        #     affine = get_affine(ds)
+        #     if ds.gridtype == "structured":
+        #         x, y = affine * np.meshgrid(x, y)
+        #     else:
+        #         x, y = affine * (x, y)
```
OnnoEbbens, line 508: can be removed?

nlmod/visualise/plots.py
```diff
@@ -355,20 +356,22 @@ def plot_vertex_array(da, vertices, ax=None, gridkwargs=None, **kwargs):
     return ax
 
 
-def da(da, ds=None, ax=None, **kwargs):
+def da(da, ds=None, ax=None, rotated=False, **kwargs):
     """
-    PLot an xarray DataArray
+    Plot an xarray DataArray, suning information from the model Dataset ds
```
OnnoEbbens, line 361: suning?

OnnoEbbens (commented):
nice features! Only some easy-to-fix comments from me.

OnnoEbbens:
When I ran the example notebooks I found some other issues and solved them. I also removed the fit_extent_to_regis function. I think everything works fine without.

nlmod/mdims/mgrid.py
rubencalje: Sometimes you want to have a modelgrid-object in model coordinates. When plotting with nlmod.plot.da() the method get_vertices() is called, which generates a modelgrid object with the default value for the argument 'rotated' set to False. But I think we can code this more elegantly and faster in the future, by skipping the modelgrid object and generating vertices streight from model_ds. Right not get_vertices() can take a while...

nlmod/mdims/mgrid.py
rubencalje: Fine. I never know which Exception I should pick, and therefore just take the most basic one.

nlmod/mdims/resample.py
rubencalje: Will add it. It is a probably a required dependency of rioxarray and ratserio as well.

Issue #65, opened 2022-07-14
Update idomain with info from layer thickness upon creation of DISV


bdestombe:
https://github.com/ArtesiaWater/nlmod/blob/ce8479333b580652150e32f006fbb5823f7944ea/nlmod/mfpackages/mfpackages.py#L161

```python
topbot = np.concatenate([model_ds.top.data[np.newaxis], model_ds.bot.data], axis=0)
thickness = -1 * np.diff(topbot, axis=0)
idomain = xr.where(thickness < 1e-5, -1, model_ds["idomain"])
```

bdestombe:
```python
    def update_idomain_(idomain, botm, top):
        topbotm = np.concatenate([top.data[None], botm], axis=0)
        thickness = -1 * np.diff(topbotm, axis=0)

        # set idomain to -1 where thickness < 1e-5
        idomain = xr.where(thickness < 1e-5, -1, idomain)

        # if the top cell is inactive idomain is 0, otherwise it is -1
        idomain = xr.concat(
            (
                xr.where(idomain[0] == -1, 0, idomain[0]),
                xr.where((idomain[:-1] == 0).data & (idomain[1:] == -1).data, 0, idomain[1:])
            ),
            dim='layer'
        )

        return idomain
```

OnnoEbbens:
Yes I agree, this is not the best place to modify the Idomain.

dbrakenhoff:
Can be closed?

dbrakenhoff:
check topbot method output (why return 2 items)?

PR #98, opened 2022-10-27
remove top3d from calculate_thickness, fix for #65

OnnoEbbens:


nlmod/mdims/mlayers.py
```diff
@@ -92,7 +86,13 @@ def layer_split_top_bot(ds, split_dict, layer="layer", top="top", bot="botm"):
     """
 
     # calculate thickness
-    thickness, top3d = calculate_thickness(ds, top=top, bot=bot)
+    thickness = calculate_thickness(ds, top=top, bot=bot)
+
+    # check if top is 2d or 3d
+    if ds[top].ndim == ds[bot].ndim:
```
dbrakenhoff, line 92: this can be simplified

dbrakenhoff (changes requested):
Looks good, just the one simplifcation for checking if top is 3d

PR #118, opened 2022-10-31
Fix #93, use surface drain resistance

### LINKED ISSUE #93: default value for surface drain conductance
The surface drain conductance gets a default value of 1000, like for the AHN.

That does not seem right when working with smaller cells, especially with refinements around rivers. Is it not more practical to define a default resistance for the surface level drain?

dbrakenhoff:


rubencalje:
> Test fail is AHN connection error.
> 
> @rubencalje I added the area data variable with cell areas to `to_model_ds()`, not sure if the name is clear enough, and whether we want to add that or just calculate it when we need it? What are your thoughts on that?

I think adding the 'area' variable is a good idea. It is already added to the model Dataset for vertex grids, so now it is always available.

PR #141, opened 2023-02-13
Improve GeoTOP code

rubencalje:
This pull request improves the way GeoTOP is used to generate geohydrological properties for groundwater models. Some improvements are:
- geohydrological properties (kh and kv) can now be calculated from a combination of lithok and strat as well.
- geohydrological properties (kh and kv) can now also be calculated from the probabilities of each lithoclass.
- GeoTOP data can be aggregated to any layer model in a model Dataset using nlmod.read.aggregate_to_ds()
- GeoTOP data can now be used to replace any layer of REGIS in get_combined_layer
- the GeoTOP code has been simplified
-  a notebook is added to demonstrate the first three points on this list

### DIFF
--- a/nlmod/read/geotop.py
+++ b/nlmod/read/geotop.py
@@ -66,27 +66,14 @@ def get_kh_kv_table(kind="Brabant"):
 
 
 @cache.cache_netcdf
-def get_geotop(extent, regis_ds, regis_layer="HLc"):
+def get_geotop(extent):
     """get a model layer dataset for modflow from geotop within a certain
     extent and grid.
-
-    if regis_ds and regis_layer are defined the geotop model is only created
-    to replace this regis_layer in a regis layer model.
-
 
     Parameters
     ----------
     extent : list, tuple or np.array
         desired model extent (xmin, xmax, ymin, ymax)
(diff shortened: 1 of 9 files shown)


nlmod/read/geotop.py
```diff
@@ -228,174 +203,403 @@ def get_top_bot_from_geo_eenheid(geotop_ds_raw, geo_eenheid_translate_df):
...
-    ds[kv] = xr.concat(kv_ar, pd.Index(ds.layer, name="layer"))
+    if add_kd_and_c:
+        ds[kd] = xr.concat(kD_ar, ds.layer)
+        ds[c] = xr.concat(c_ar, ds.layer)
+    ds[kh] = xr.concat(kh_ar, ds.layer)
+    ds[kv] = xr.concat(kv_ar, ds.layer)
+    return ds
+
+
+def plot_cross_section(
```
OnnoEbbens, line 566: I think this should be moved to the plot.py module (with a docstring)

OnnoEbbens (approved):

OnnoEbbens:
Good job! I only added one comment and already approved this PR

PR #159, opened 2023-04-07
Add Meteobase ASCII read function

martinvonk:
This PR adds the option to read rasterdata zip files containing ASCII  data from [Meteobase](https://www.meteobase.nl/?tb=rasterdata&dp=rasterdata&dp_sub=introductie). Unfortunately, data can't be downloaded using an _api_ so a zip file has to be downloaded manually.


martinvonk:
Should be all good now @rubencalje

rubencalje (approved):
Great addition!

PR #169, opened 2023-05-01
Create nlmod Datasets from flopy modelgrids

dbrakenhoff:
Add functionality to create datasets from flopy modelgrids.

Usage:
```python

gwf = sim.get_model("model")  # given some existing flopy MFSimulation + model

ds = nlmod.grid.modelgrid_to_ds(gwf.modelgrid)  # convert to Dataset

# add data from flopy
ds["k"] = ("layer", "cell2d"), gwf.npf.k.array
```

Perhaps this can serve as a base for future more complete `nlmod.read.flopy` functionality?


rubencalje:
> > @rubencalje, any idea what's causing the test failure?
> 
> I have no idea. Maybe fiona is updated. I would ignore it for now.

Geopandas updated to version 0.13.0 on may 6, which is probably the reason of the failure. I will look for a fix.

rubencalje:
> @rubencalje, any idea what's causing the test failure?

I have no idea. Maybe fiona is updated. I would ignore it for now.

> > > @rubencalje, any idea what's causing the test failure?
> > 
> > 
> > I have no idea. Maybe fiona is updated. I would ignore it for now.
> 
> Geopandas updated to version 0.13.0 on may 6, which is probably the reason of the failure. I will look for a fix.

I fixed this in my pull-request #174, so you can ignore it here.

rubencalje:
I have two comments:
- should we make `get_structured_grid_ds` and `get_vertex_grid_ds` internal methods, or do we want users to call them directly?
- should `cell2d` not be an argument of `get_vertex_grid_ds`? The handling of `cell2d` is performed in `modelgrid_to_ds`, so `get_vertex_grid_ds` does not create a fully functioning model-ds, where the vertices are connected to the cells.

dbrakenhoff:
@rubencalje Good suggestions, I'll push those directly to dev as soon as i manage to merge #170.

dbrakenhoff:
@rubencalje Good suggestions, I'll push those directly to dev as soon as i manage to merge #170.

PR #178, opened 2023-05-15
Improve getting time coordinates from groundwater model objects

dbrakenhoff:
- Build time coordinate from flopy groundwater model object, usage: `nlmod.time.ds_time_from_model(model)` 
- Use new function to obtain time coordinate when reading heads/concentrations using groundwater model object.
- Add function to get budget for specific record as DataArray

### DIFF
--- a/nlmod/dims/time.py
+++ b/nlmod/dims/time.py
@@ -236,8 +236,10 @@ def ds_time_from_model(gwf):
     start_datetime = gwf.simulation.get_package("TDIS").start_date_time.data
 
     if start_datetime is not None:
+        time_units = gwf.dimensions.simulation_time.get_time_units()
         dt = pd.to_timedelta(
-            np.cumsum(gwf.dimensions.simulation_time.get_perioddata()["perlen"]), "D"
+            np.cumsum(gwf.dimensions.simulation_time.get_perioddata()["perlen"]),
+            time_units,
         )
         times = pd.Timestamp(start_datetime) + dt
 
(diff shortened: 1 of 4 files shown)


nlmod/dims/time.py
```diff
@@ -216,3 +217,36 @@ def estimate_nstp(
...
+    -------
+    IndexVariable
+        time coordinate for xarray data-array or dataset
+    """
+
+    start_datetime = gwf.simulation.get_package("TDIS").start_date_time.data
+
+    if start_datetime is not None:
+        dt = pd.to_timedelta(
+            np.cumsum(gwf.dimensions.simulation_time.get_perioddata()["perlen"]), "D"
```
OnnoEbbens, line 240: Here you use "D" as time unit

nlmod/dims/time.py
```diff
@@ -216,3 +217,36 @@ def estimate_nstp(
...
+        )
+        times = pd.Timestamp(start_datetime) + dt
+
+    else:
+        times = np.cumsum(gwf.dimensions.simulation_time.get_perioddata()["perlen"])
+
+    time = IndexVariable(["time"], times)
+    time.attrs["time_units"] = gwf.dimensions.simulation_time.get_time_units()
```
OnnoEbbens, line 248: here you get the time unit from the gwf

nlmod/gwf/output.py
```diff
@@ -129,7 +221,7 @@ def _get_hds(ds=None, gwf=None, fname_hds=None):
     return headobj
 
 
-def _get_cbc(ds=None, gwf=None, fname_cbc=None):
+def _get_cbf(ds=None, gwf=None, fname_cbc=None):
```
OnnoEbbens, line 224: I think it should be cbc? If not the argument `fname_cbc` should be changed as well?

OnnoEbbens (commented):
Some minor things to fix and then it is good to go

PR #182, opened 2023-05-24
Small fixes to Meteobase

bdestombe:


martinvonk:
Thanks for the enhancements. I made a small few changes to them to make sure the (meta)data is actually available. This does always seem to be the case with the MeteoBase data (see #173).

PR #184, opened 2023-06-01
add test for lake

OnnoEbbens:
Added a test for the lake package. The test covers a simple case with 2 lakes, evaporation and one like overflowing in the other.


bdestombe:
Great worrk Onno!

Similarly to time_conversion, we should check the length unit is meters to deal with the dimension specific constants of Manning's eq. length_conversion should be fixed to 1 as we model in meters.

The manual says the following:
```length conversion—real value that is used in converting outlet flow terms that use Manning’s equation or gravitational acceleration to consistent length units. LENGTH CONVERSION should be set
to 3.28081, 1.0, and 100.0 when using length units (LENGTH UNITS) of feet, meters, 
or centimeters in the simulation, respectively. LENGTH CONVERSION does not 
need to be specified if no lake outlets are specified or LENGTH UNITS are meters.
```

Other than that, looks good to me!

rubencalje:
> Great worrk Onno!
> 
> Similarly to time_conversion, we should check the length unit is meters to deal with the dimension specific constants of Manning's eq. length_conversion should be fixed to 1 as we model in meters.
> 
> The manual says the following:
> 
> ```
> to 3.28081, 1.0, and 100.0 when using length units (LENGTH UNITS) of feet, meters, 
> or centime�ters in the simulation, respectively. LENGTH CONVERSION does not 
> need to be specified if no lake outlets are specified or LENGTH UNITS are meters.
> ```
> 
> Other than that, looks good to me!

Good remark. We do not add a length unit to ds right now, so I added a todo-comment, so we can merge this Pull Request now.

PR #185, opened 2023-06-05
Flip meteobase data and bug in loading data

bdestombe:


martinvonk:
Hi @bdestombe, 

Thanks for the PR. Can you tell me which _bug_ this solves? And why the data has to be flipped?

Also, on the nodata default argument; the DataArray metadata contains the _NODATA_value_ and if this is replaced by np.nan this value is not valid anymore. So I like the default argument to be False such that filling this np.nan is a conscious decision of the user. Or we have to replace the _NODATA_value_ in the metadata.

bdestombe:
Hi Martin,
The image was flipped vertically, as was clearly visible in the evapotranpiration product.

Nans were invented exactly for this purpose, so why not use them? The nodata value is used when working with integers. And some file formats require working with integers such that nodata values are inherently required. That is not the case here, as all the products contain floats and xarray manages nans quite well.

When exported to tiffs the nans get correctly propagated otherwise you'd manually have to assign the no data value via rioxarray. Furthermore, the xarray plotting utilities work well with nans.

But you are right, the nodata_value now has no further meaning. I'll remove it in the PR.

Thanks for your feedback!

martinvonk:
> The image was flipped vertically, as was clearly visible in the evapotranpiration product.

Is this also the case in the other products? Or is this a problem with MeteoBase?

bdestombe:
> > The image was flipped vertically, as was clearly visible in the evapotranpiration product.
> 
> Is this also the case with the other products? Or is this a problem with MeteoBase?

Good question.. I am not sure how to check this.. Any suggestion?

bdestombe:
Both evapotranspiration and verdampingstekort are now orientated correctly. I'm not sure how to check precepitation. And Manning en Penman only contain nodata values on the dataset that I have at hand.

martinvonk:
We could maybe compare the precipitation dataset to the KNMI radar dataset (#166). See how that one compares to the flipped version. 

The dataset for meteobase does not contain NODATA for the Makkink and Penman evaporation. https://github.com/ArtesiaWater/nlmod/blob/main/tests/data/Meteobase_ASCII_test.zip

bdestombe:
Ah indeed! It shows data, but its too small with too little resolution to distinguish features.

The produced geotiffs of your test data are attached.
[out_test.zip](https://github.com/ArtesiaWater/nlmod/files/11652877/out_test.zip)

martinvonk:
I think it is fine now so we can merge this. 
I now flip the data when constructing the y-coordinates from the metadata, instead of flipping the data when reading the ASCII. 

I wasn't able to confirm the precipitation dataset with the KNMI data. KNMI uses another stereographic projection and I don't have time for that at the moment.

PR #188, opened 2023-06-05
Handling data when building flopy packages from model dataset

dbrakenhoff:
This PR aims to inform users how nlmod is dealing with data when users build flopy packages.

In terms of variables we're trying to stick to flopy default names as much as possible. This is mostly the case, though some differences might remain. To distinguish datasets between different packages the convention we're using is `<package-abbreviation>_<variable-name>`.

Basically, there are two options:
1. variables are passed as strings, and point to an array in the dataset
2. variables are passed as arrays (data arrays, or classic numpy arrays, or constants).  

Option 2 will have precedence over option 1, so passing your own data-array will always override any stored dataset (with the same name).

Option 1:
- notify user data is read from dataset (debug level)
- notify user if dataset does not exist and return None (warning)

Option 2:
- if a data array is passed, notify user if data with the same name is stored in dataset (info)
- if any other datatype, just use passed value (e.g. a numpy array), (no message)

The function `_get_value_from_ds_datavar` takes care of this behavior for data variables. The function `_get_value_from_ds_attr` does something similar for attributes. Both these functions are tested in `test_003_mfpackages.py`, which can also be helpful in understanding how to call these methods. 

Other things worth mentioning:
- deprecated `da_name` keyword arguments for `nlmod.gwf.ghb` and `nlmod.gwf.drn` packages
- deprecated `chd`  keyword argument, renamed to `mask` in `nlmod.gwf.chd`
- rename "_peil" data variable to "_stage" for RWS surface water data


dbrakenhoff:
CI error is unrelated to this PR (something about data from Wetterskip Fryslan), so this is ready for review.

PR #192, opened 2023-06-07
Update to 0.6.0

rubencalje:


dbrakenhoff:
Lots of stuff here! Seems like this update might deserve more than a minor version bump? Perhaps 0.6.0?

PR #194, opened 2023-06-21
Improve reading model output

dbrakenhoff:
Use dask.delayed to load model output. This avoids reading all data into memory if `delayed=True`. Optionally chunk data array for doing memory efficient calculations on large data arrays. 

Adds two kwargs to output methods:
- `delayed`, if True, do not load data into memory, default is False
- `chunked`, if True, chunk data array using `da.chunk("auto")`, default is False

Default behavior is same as before, for memory intensive output, use `delayed=True`, and optionally `chunked=True`, e.g.:

```python
heads_orig = nlmod.gwf.output.get_heads_da(ds)  # read all data into memory
heads_delayed = nlmod.gwf.output.get_heads_da(ds, delayed=True)  # memory efficient
heads_chunked = nlmod.gwf.output.get_heads_da(ds, delayed=True, chunked=True)  # chunked
```


nlmod/gwf/output.py
```diff
@@ -40,15 +61,38 @@ def get_heads_da(ds=None, gwf=None, fname_heads=None, fname_hds=None):
...
+        delayed=delayed,
+        chunked=chunked,
+    )
     head_da.attrs["units"] = "m NAP"
     return head_da
 
 
-def get_budget_da(text, ds=None, gwf=None, fname_cbc=None, kstpkper=None):
+def _get_cbc(ds=None, gwf=None, fname_cbc=None):
+    msg = "Load the budgets using either the ds or the gwf"
```
OnnoEbbens, line 80: docstring

nlmod/mfoutput.py
```diff
@@ -42,31 +50,42 @@ def _get_output_da(reader_func, ds=None, gwf_or_gwt=None, fname=None, **kwargs):
...
-    arr[arr == hnoflo] = np.nan
+    stacked_arr = dask.array.stack(result)
 
     if gwf_or_gwt is not None:
         gridtype = gwf_or_gwt.modelgrid.grid_type
     else:
         gridtype = ds.gridtype
 
     if gridtype == "vertex":
+        layers = np.arange(stacked_arr.shape[1])
```
OnnoEbbens, line 72: If a dataset is given I think it is better to use the layers from the dataset. Otherwise you might overwrite the layers dimension of the dataset.

nlmod/mfoutput.py
```diff
@@ -42,31 +50,42 @@ def _get_output_da(reader_func, ds=None, gwf_or_gwt=None, fname=None, **kwargs):
...
+        # stacked arr is 4d
         da = xr.DataArray(
-            data=arr[:, :, 0],
-            dims=("time", "layer", "icell2d"),
-            coords={},
+            data=stacked_arr,
+            dims=("time", "layer", "_drop", "icell2d"),
+            coords=coords,
         )
+        # drop unnecessary dimension
```
OnnoEbbens, line 87: Why not slice the array to remove the unnecessary dimension as was done before?

OnnoEbbens (approved):
Good job!

I was thinking if we should make this functions cache-able using the `cache_netcdf` decorator but I think it is too hard (and to little gains) because the `cache_netcdf` requires a dataset as input and for this function it is optional. However if these functions do become annoyingly slow we could evaluate again because it is certainly possible to use the cache here.

nlmod/gwf/output.py
dbrakenhoff: Added

nlmod/gwf/output.py
```diff
@@ -14,7 +14,24 @@
 logger = logging.getLogger(__name__)
 
 
-def get_heads_da(ds=None, gwf=None, fname_heads=None, fname_hds=None):
+def _get_heads(ds=None, gwf=None, fname_hds=None):
```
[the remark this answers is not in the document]
dbrakenhoff: Added.

nlmod/mfoutput.py
dbrakenhoff: good suggestion! I slice the stacked_array now.

nlmod/gwf/output.py
```diff
@@ -84,7 +84,7 @@ def get_heads_da(
         Xarray dataset with model data.
     gwf : flopy ModflowGwf
         Flopy groundwaterflow object.
-    fname : path, optional, Deprecated
+    fname : path, optional
         Instead of loading the binary heads file corresponding to ds or gwf
         load the heads from
```
OnnoEbbens, line 89: I think something is missing here..

OnnoEbbens (approved):

nlmod/dims/grid.py
```diff
@@ -224,6 +224,47 @@ def modelgrid_to_ds(mg):
...
+        layers = np.arange(mg.nlay)
+        y = mg.ycellcenters
+        x = mg.xcellcenters
+        coords = {"layer": layers, "y": ("icell2d", y), "x": ("icell2d", x)}
+        dims = ("layer", "icell2d")
+    else:
+        raise ValueError(f"grid type '{mg.grid_type}' not supported.")
+    return dims, coords
+
+
```
OnnoEbbens, line 267: sort of similar to `dims.grid.modelgrid_to_ds`, If you have time I would suggest moving this function to dims.grid and combine with the current functions

nlmod/plot/flopy.py
```diff
@@ -111,6 +111,72 @@ def map_array(
...
+    alpha=1.0,
+    labels=True,
+    label_kwargs={},
+    plot_grid=False,
+    add_to_plot=None,
+    backgroundmap=False,
+    figsize=None,
+    animate=False,
+    **kwargs,
+):
```
OnnoEbbens, line 134: Can you add a docstring

nlmod/dims/grid.py
dbrakenhoff: Yea, I first wanted to use the `modelgrid_to_ds` function but creating a full dataset with tops and bottoms seemed a bit excessive.

Maybe we could `use get_dims_coords_from_modelgrid` in `modelgrid_to_ds` somehow, but I didn't see a very obvious way right away.

OnnoEbbens (approved):

Issue #124, opened 2022-11-02
input flexibility set_ds_time rt perioddata


tomvansteijn:
Currently the input function `set_ds_time` does not allow for the full flexibilty of the perioddata variable of the Flopy TDIS interface. 

Usage of the input variable `time` forces timesteps of 1 day, which is not always necessary. Furthermore the timestamp multiplication factor `tsmult` can only be applied as a constant for all stress periods. It would be useful if the number of timesteps (ntsp) and the multiplication factor `tsmult` could be specified in conjunction with the `perlen` and the `time` variable.

In some cases it is more practical to specify a start date and the stress period data in the same format as required by Flopy TDIS. In other cases just the `time` variable should be sufficient, in my experience.

rubencalje:
Thanks for the suggestions. Some of the issues are allready implemented (but documentation can probably be improved). Here are some comments:

> Usage of the input variable time forces timesteps of 1 day, which is not always necessary. 

This is not true. For example, see the following example for yearly stress-period lengths, with one timestep per stress-period (nstp=1):

```
import pandas as pd
import nlmod
ds = nlmod.get_ds([0, 3000, 0, 2000])
ds = nlmod.time.set_ds_time(ds, time=pd.date_range('2010', '2020', freq='YS'))
ds.time
```
gives
```
<xarray.DataArray 'time' (time: 11)>
array(['2010-01-01T00:00:00.000000000', '2011-01-01T00:00:00.000000000',
       '2012-01-01T00:00:00.000000000', '2013-01-01T00:00:00.000000000',
       '2014-01-01T00:00:00.000000000', '2015-01-01T00:00:00.000000000',
       '2016-01-01T00:00:00.000000000', '2017-01-01T00:00:00.000000000',
       '2018-01-01T00:00:00.000000000', '2019-01-01T00:00:00.000000000',
       '2020-01-01T00:00:00.000000000'], dtype='datetime64[ns]')
Coordinates:
  * time     (time) datetime64[ns] 2010-01-01 2011-01-01 ... 2020-01-01
Attributes:
    time_units:    DAYS
    start:         2000-01-02 00:00:00
    nstp:          1
    tsmult:        1.0
    steady_start:  1
    steady_state:  0
```

> Furthermore the timestamp multiplication factor tsmult can only be applied as a constant for all stress periods. It would be useful if the number of timesteps (ntsp) and the multiplication factor tsmult could be specified in conjunction with the perlen and the time variable.

[...]

PR #257, opened 2023-08-28
Improve time discretization

dbrakenhoff:
This PR aims to simplify and improve the handling of time discretization in nlmod.

To add time index to your model dataset use `nlmod.time.set_ds_time()`\*.

_*NOTE: this function replaces the old function with the same name, but works differently, so this is a minor breaking change. The old function can still be called through `nlmod.time.set_ds_time_deprecated()` for now._

The arguments to `nlmod.time.set_ds_time()` are:
- `time` is an array-like with the timestamps at the end of each stress period (note that this does not include the starting time of the model)
- `start` is the start datetime of the model
- `steady` is an array-like indicating whether a stress period is steady-state (True or 1) or transient (False or 0). The default is steady-state for all timesteps.
- `nstp` and `tsmult` are optional arguments to indicate the number of time steps per stress period and the timestep multiplier. These are used when writing the Modflow6 TDIS file.

```python
# basic call signature
ds = nlmod.time.set_ds_time(ds, time, start, steady)

# set ds time with array of elapsed time periods
ds = nlmod.time.set_ds_time(ds, [100, 200], "2020-01-01", steady=[1, 0])

# set ds time with perlen
perlen = [100, 100]
ds = nlmod.time.set_ds_time(ds, np.cumsum(perlen), "2020-01-01", steady=[1, 0])  # note the cumsum

# set ds time with dates
ds = nlmod.time.set_ds_time(ds, ["2020-04-10", "2020-07-19"], "2020-01-01", steady=[1, 0])
```

A function has been added to convert flopy TDIS period data settings (`perlen`, `nstp`, `tsmult`) into a time index:
```python
# for getting the time index for all time steps
ds = nlmod.time.ds_time_idx_from_tdis_settings(start, perlen, nstp, tsmult)

# for getting the time index at the end of each stress period, omit the nstp and tsmult keyword arguments
ds = nlmod.time.ds_time_idx_from_tdis_settings(start, perlen)
```

EDIT: Some additional comments on this PR after @OnnoEbbens review:

The variables `steady`, `nstp` and `tsmult` are now stored as data variables in the model dataset with dimension time. This should make them more visible and easier to modify for users as well. 

A check is included whether the first entry of the time array (after conversion to a datetime object) is equal to or lies before the start datetime. This should help users figure out if they've made a mistake when assigning these variables.


nlmod/dims/time.py
```diff
@@ -127,6 +136,99 @@ def set_ds_time(
...
+    # set steady
+    if isinstance(steady, bool):
+        steady = steady * np.ones(len(time))
+
+    ds = ds.assign_coords(coords={"time": time})
+    if time_units == "D":
+        time_units = "DAYS"
+    ds.time.attrs["time_units"] = time_units
+    ds.time.attrs["start"] = str(start)
+    ds.time.attrs["steady"] = steady
```
OnnoEbbens, line 202: I think steady is always a boolean numpy array. We could also add this as a DataVariable to the Dataset with dimension time.

nlmod/dims/time.py
```diff
@@ -127,6 +136,99 @@ def set_ds_time(
...
+    ds.time.attrs["nstp"] = nstp
+    ds.time.attrs["tsmult"] = tsmult
+    return ds
+
+
+def ds_time_idx_from_tdis_settings(start, perlen, nstp=1, tsmult=1.0, time_units="D"):
```
OnnoEbbens, line 208: docstring?

nlmod/sim/sim.py
```diff
@@ -92,15 +92,21 @@ def get_tdis_perioddata(ds):
...
-        nstp = [ds.time.nstp] * len(perlen)
+    nstp = util._get_value_from_ds_datavar(
+        ds, "nstp", nstp, return_da=False, warn=False
+    )
+    nstp = util._get_value_from_ds_attr(ds.time, "nstp", nstp)
+
+    if isinstance(nstp, (int, np.integer)):
+        nstp = [nstp] * len(perlen)
+
+    nstp = util._get_value_from_ds_datavar(
```
OnnoEbbens, line 103: I think this one should be `tsmult` instead of `nstp`

nlmod/sim/sim.py
```diff
@@ -64,7 +64,7 @@ def write_and_run(sim, ds, write_ds=True, script_path=None, silent=False):
     ds.attrs["model_ran_on"] = dt.datetime.now().strftime("%Y%m%d_%H:%M:%S")
 
 
-def get_tdis_perioddata(ds):
+def get_tdis_perioddata(ds, nstp=None, tsmult=None):
```
OnnoEbbens, line 67: I prefer not to have `nstp` and `tsmult` as keyword arguments here because than you can end up with different settings in your dataset and your tdis.

tests/test_001_model.py
```diff
@@ -18,20 +20,17 @@ def test_model_directories(tmpdir):
...
-        ds,
-        start_time="2015-1-1",
-        steady_state=False,
-        steady_start=True,
-        transient_timesteps=10,
-    )
+    nper = 11
+    time = pd.date_range(start="2015-1-1", periods=nper, freq="D")
+    steady = np.zeros(nper)
+    ds = nlmod.time.set_ds_time(ds, time, start="2015-1-1", steady=steady)
```
OnnoEbbens, line 33: I think the `start` should not be the same as the first item in `time` since `time` is the time at the end of each period.

tests/test_001_model.py
```diff
@@ -354,3 +359,8 @@ def _check_tmpdir(tmpdir):
         return tmpdir
     else:
         return str(tmpdir)
+
+
+if __name__ == "__main__":
+    test_create_sea_model_grid_only(tmpdir)
+    test_create_sea_model(tmpdir)
```
OnnoEbbens, line 366: This can be removed

OnnoEbbens (changes requested):
Good job! some comments from me nothing big.

nlmod/dims/time.py
dbrakenhoff: Good idea, I like that.

OnnoEbbens (approved):

rubencalje:
I liked the original method because of the following default values:
- the first period would be steady state, following periods would be transient
- the first steady state period had a length of 3652 days (10 years)

Multiple steady state stress periods is probably only useful in transport calculations, so I think a transient run for multiple stress periods should be the default. With the above default values the starting conditions of the transient run are determined by a steady-state run of 10 years. To get the same behavior of the new method would require a few lines of code from the user.

Also, the method `set_ds_time_deprecated` is not that useful, as it does not store the time settings in the right way for the rest of nlmod anymore.

Therefore, I would be more in favor of fixing/improving the original method, in which we:
- deprecate some input-parameters like `perlen` and `transient_timesteps` (with a deprecation message, or just an appropriate exception message)
- keep 'steady_start' with a default value of True (like it is now)
- replace `steady_state` by `steady`, with the same default value of False, and use the proposed method in this PR to store the steady_state stress periods in ds['steady'].

dbrakenhoff:
Thanks for the review! I'm a bit biased, but i find the old code almost unreadable, so I'd be in favor of modifying this PR further to modify the defaults to what you suggest.

I agree we most often want to default to one steady state period and then subsequent transient periods. That's something we can implement with an extra keyword argument? And then explicitly entering the 10 year period isn't that much effort on the part of the user, just subtract 10 years from `time[0]` or just entering a year. This has the added advantage of being much more explicit :)?

So to reproduce current behavior it would become something like this (with these values being default for the steady keyword args):

`nlmod.time.set_ds_time([3650, 100, 100], start="2010", steady=False, steady_start=True)`

rubencalje:
The old code was complicated, because it needed to support `perlen` (which is mentioned as a positive in issue #124) and `transient_timesteps`. Let's forget the old code, and work on the lines in this PR.

It is a good idea to set the length of the steady-state start period more explicit. I can work with your proposed code:
`time = pd.date_range("2020-01-01", "2025-01-01", "MS")`
`nlmod.time.set_ds_time(time, start="2010", steady_start=True)`

dbrakenhoff:
Alright, sounds good!

As for perlen, its still supported but you have to take the cumulative sum in this new function.

```python
# set ds time with perlen
perlen = [100, 100]
ds = nlmod.time.set_ds_time(ds, np.cumsum(perlen), start="2020")  # note the cumsum
```

rubencalje:
I added perlen, also because it is requested in issue #124, and set a default value for `start`.

I would remove the log-message at the start of the method, as this will get annoying, I think.

dbrakenhoff:
Nice! My thoughts on these additions:

- Adding a `perlen` option is nice! I do think it should be treated exactly the same as the time option (so type should always be array-like. Or we modify time to also support single values (but I'm more for option 1). 
- I'm not a big fan of the `steady` default change though. I wanted users to explicitly provide this value because it makes it very obvious what the time period you're modeling is going to be. I do like the option of allowing a number and interpreting that as a Timedelta relative to the first entry in `time`. This does add a layer of difficulty if time is passed as elapsed time, or if `perlen` is provided instead of `time`, in which case there is no reference point, but I saw you already included a check for that.  

So in short, I'm for perlen always array-like, and getting rid of the steady defaults and forcing the user to pick something. 

Perhaps we can get @OnnoEbbens to weigh in and help us pick a way to go :).

rubencalje (approved):

rubencalje:
I fixed some bugs relating to previous changes and the new pandas version. They are not really related to this pull request, but were needed to test all the notebooks for the time-related changes. I think this PR is ready to be merged.

docs/examples/00_model_from_scratch.ipynb
dbrakenhoff, line 1: This notebook uses `start=1`, is this something that should work, or should this be a datetime? 

`nlmod.time.set_ds_time(ds, time=pd.Timestamp.today(), start=1)`

dbrakenhoff: Ah, I didn't pull in the most recent changes and was getting an error. Sorry.

dbrakenhoff (commented):
I have one fix for nb 11, for the rest all good to go!

Issue #260, opened 2023-09-11
reinstate `fit_extent_to_regis` under a different name


OnnoEbbens:
The `fit_extent_to_regis` function resized a random extent in such a way that there will be an integer number of columns and cells given a certain cellsize. This is useful in case you want to quickly create a grid from a random extent.

I would propose to reinstate the function under the name `snap_extent` in de `grid` module.

The original `fit_extent_to_regis` is:

```
def fit_extent_to_regis(extent, delr, delc, cs_regis=100.,
                        verbose=False):
    """
    redifine extent and calculate the number of rows and columns.

    The extent will be redefined so that the borders os the grid (xmin, xmax, 
    ymin, ymax) correspond with the borders of the regis grid.

    Parameters
    ----------
    extent : list, tuple or np.array
        original extent (xmin, xmax, ymin, ymax)
    delr : int or float,
        cell size along rows, equal to dx
    delc : int or float,
        cell size along columns, equal to dy
    cs_regis : int or float, optional
        cell size of regis grid. The default is 100..

    Returns
    -------
    extent : list, tuple or np.array
        adjusted extent
    nrow : int
        number of rows.
    ncol : int
        number of columns.

    """
    extent = extent.copy()
    
    if verbose:
        print(f'redefining current extent: {extent}, fit to regis raster')

    for d in [delr, delc]:
        if float(d) not in [10., 20., 25., 50., 100., 200., 400., 500., 800.]:
            print(f'you probably cannot run the model with this '
                  f'cellsize -> {delc, delr}')

    # if extents ends with 50 do nothing, otherwise rescale extent to fit regis
    if extent[0] % cs_regis == 0 or not extent[0] % (0.5 * cs_regis) == 0:
        extent[0] -= extent[0] % 100
        extent[0] = extent[0] - 0.5 * cs_regis
    # get number of columns
    ncol = int(np.ceil((extent[1] - extent[0]) / delr))
    extent[1] = extent[0] + (ncol * delr)  # round x1 up to close grid

    # round y0 down to next 50 necessary for regis
    if extent[2] % cs_regis == 0 or not extent[2] % (0.5 * cs_regis) == 0:
        extent[2] -= extent[2] % 100
        extent[2] = extent[2] - 0.5 * cs_regis
    nrow = int(np.ceil((extent[3] - extent[2]) / delc))  # get number of rows
    extent[3] = extent[2] + (nrow * delc)  # round y1 up to close grid

    if verbose:
        print(
[...]

PR #266, opened 2023-10-10
fix for #260

OnnoEbbens:


rubencalje:
Hmm, I see from the tests:
```
extent = (0.22, 1056.12, 7.43, 1101.567)
new_extent = nlmod.dims.snap_extent(extent, 10, 20)
assert new_extent == [-5.0, 1065.0, -10.0, 1110.0]
```

So the **center** of cells within the new extent is snapped to the closest multiple of delr/delc. I am more in favor of snapping the **edge** of cells to the closest multiple of delr/delc, like:

```
extent = (0.22, 1056.12, 7.43, 1101.567)
new_extent = nlmod.dims.snap_extent(extent, 10, 20)
assert new_extent == [0.0, 1060.0, 0.0, 1120.0]
```

Regis is also defined in intervals of 100 m, so from x = 100.000 until 100.100, x = 100.100 until 100.200 and so on. With your approach half of a cell would be in one regis-cell, and the other half would be in another one (if I am not mistaken).

OnnoEbbens:
I agree that makes a lot more sense

OnnoEbbens:
Two of the tests for the knmi data platform fail that are not related to this, I created an issue for this (https://github.com/ArtesiaWater/nlmod/issues/272) and will merge this PR.

PR #269, opened 2023-10-13
Incomplete knmi ts breaks get_recharge function

bdestombe:
Solution provided by Davíd.

Missing values are filled with nearby time series by setting fill_missing_obs to True

### DIFF
--- a/nlmod/read/knmi.py
+++ b/nlmod/read/knmi.py
@@ -277,11 +277,11 @@ def get_knmi_at_locations(ds, start="2010", end=None, most_common_station=False)
 
     # get knmi data stations closest to any grid cell
     oc_knmi_prec = hpd.ObsCollection.from_knmi(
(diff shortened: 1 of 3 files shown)


rubencalje:
I updated the fill_missing_obs functionality in hydropandas this week. It does not check if de Bilt has measurements anymore, if this is not needed (when no data needs to be filled). So we maybe need to increase the hydropandas version in the next nlmod release (if the new hydropandas has been released).

rubencalje:
Also, evaporation-measurements were not filled in Hydropandas. This has been fixed in a PR last week: https://github.com/ArtesiaWater/hydropandas/pull/157

OnnoEbbens:
I've updated hydropandas and pinned the version in the requirements of nlmod. I would like to merge this PR asap so we can transfer nlmod to a new organisation without any open PR's

Issue #261, opened 2023-09-13
Wrong coordinates in heads-DataArray from nlmod.gwf.get_heads_da(ds) for disv-grids


rubencalje:
The coordinates of the heads DataArray are wrong (too high) when using a disv-grid. These is caused by the fact that the coordinates of the vertices itself are already in real coordinates. When the translation with (xorigin, yorigin) is applied, the coordinates are almost doubled.

I think this can be solved by setting `xorigin `and `yorigin `in disv to 0 by default, unless grid rotation is used (`ds.attrs["angrot"] != 0.0`)

rubencalje:
Fixed

PR #276, opened 2023-10-23
Buisdrainage

rubencalje:
This pull request adds support for buisdrainage from the NHI-website to nlmod. Also, it fixes some small bugs in code and documentation, for example issue #261


nlmod/gwf/surface_water.py
```diff
@@ -1024,8 +1024,10 @@ def add_season_timeseries(
     package,
     summer_months=(4, 5, 6, 7, 8, 9),
     filename="season.ts",
-    seasons=("winter", "summer"),
+    seasons=None,
```
dbrakenhoff, line 1027: Why set seasons to None? The keyword summer_months sort of suggests winter/summer being the only possible two seasons?

nlmod/read/nhi.py
```diff
@@ -0,0 +1,88 @@
...
+    depth_method="mode",
+):
+    if pathname is None:
+        pathname = ds.cachedir
+    # download files if needed
+    fname_c, fname_d = download_buisdrainage(pathname)
+
+    # make sure crs is set of ds
+    if ds.rio.crs is None:
+        ds = ds.rio.write_crs(28992)
```
dbrakenhoff, line 52: should we use the EPSG definition in nlmod for this?

nlmod/gwf/surface_water.py
rubencalje: I changed this as a tuple gave a bug, `seasons` (`time_series_namerecord`) somehow needed to be a list, and it is not allowed to to add a list as an optional arggument. Therefore when `seasons` is None, I set it to `seasons = ["winter", "summer"]`. Maybe better to add two arguments, called `winter_name` and `summer_name`?

nlmod/read/nhi.py
rubencalje: Somehow this does not give the same problems as with maps (I believe, but did not thoroughly check). It might break stuff, if we change this. If other data sources are also in 28992, they do not need to be transformed, but if we change the crs of the model Dataset to our own definition, an unwanted transformation is applied. The buisdrainage data is in WGS84, so you may be right. Let's check this in the future.

Issue #275, opened 2023-10-20
Improve nlmod.layers.fill_top_and_bottom()


dbrakenhoff:
This method name is a bit vague considering the change from a 3D to 2D top is also performed. 

Maybe we can call this  to_modflow_tops_and_bottoms() which calls 2 functions, the first being fillnan_top_and_bottom() and the second make_top_2d()? Names need some work, because unstructured top is 1D...

And additionally, from Onno:

> The code to fill nan values in the top and bottom can be optimized using the xarray bfill or ffill methods.

rubencalje:
Needs top be fixed together with issue #270

Issue #270, opened 2023-10-17
Always use Modflow way of storing layers (no 3d top in layer Datasets anymore)


rubencalje:
Right now we use two definitions to store a layer/model Dataset:
1 With a 3d top, where top and bottom can contain NaN's: a layer Dataset
2 With a 2d top (or 1d for vertex grids), where top and bottom do not contain NaN's: a model Dataset

This creates a problem where methods only work for one of the above Datasets. I propose to adopt the Modflow-way of storing data for every layered Dataset, so to only store the top of the top layer, and the bottom for all layers.

This means that the method nlmod.read.get_regis() will then return something different. Regis sometimes has some holes in the layer model: the bottom of one layer is not equal to the top of the next layer. I think we should notify the user in which layer this takes place, when we remove the layer-dimension from top, using some log-messages.

This would break some peoples code, but will make life easier in the future, and saves us a lot of checking for a layer-dimension in top.

rubencalje:
This is implemented. If you want to know how large in the holes in the layer model are, you can run:
```
regis = nlmod.read.regis.get_regis(extent, drop_layer_dim_from_top=False)
nlmod.layers.remove_layer_dim_from_top(regis, return_inconsistencies=True)
```

PR #289, opened 2023-11-14
Multiple improvements

rubencalje:
This pull request does the following:
- All layer models in nlmod will no longer have a layer dimension in the top, see issue #270. So when you download regis, NaN's in top and bottom (where the layer does not occur) will be filled by overlying and underlying layers with data, and only the top of the entire model and the bottom of each layer is stored.
- It speeds up the filling of tops and bottoms (mentioned as comment by Onno in issue #275)
- `extrapolate_ds` accepts a `layer` argument, so you can use it for geotop data as well
- fix a bug in `extrapolate_ds` for 2d (non-layered) data
- add "concentration" to perioddata of buyuoncy package, to correctly account for density in modflow packages of boundary conditions (drn, ghb, riv)
- add `nlmod.plot.geotop_lithok_on_map` (tested in tests/test_002_regis_geotop.py)
- change url of webservices of Wetterkip Fryslan and Noorderzijlvest
- update test of download of surface water data from Fryslan to HHNK as this webservice of Fryslan is broken
- update modpath excecutable

So the most important change is the first: `nlmod.read.get_regis(extent)` will return a Dataset without a layer dimension in the variable `top` (which can be overridden by setting `drop_layer_dim_from_top` to False by the way). Does everybode agree?

### DIFF
--- a/nlmod/dims/layers.py
+++ b/nlmod/dims/layers.py
@@ -1380,20 +1380,17 @@ def insert_layer(ds, name, top, bot, kh=None, kv=None, copy=True):
     needs to be inserted above the existing layer, and if the top or bottom of the
     existing layer needs to be altered.
 
-    For now, this method needs a layer model with a 3d-top, like you get using the
-    method `nlmod.read.get_regis()`, and does not function for a model Dataset with a 2d
-    (structured) or 1d (vertex) top.
-
     When comparing the height of the new layer with an existing layer, there are 7
     options:
 
     1 The new layer is entirely above the existing layer: layer is added completely
     above existing layer. When the bottom of the new layer is above the top of the
     existing layer (which can happen for the first layer), this creates a gap in the
-    layer model.
+    layer model. In the returned Dataset, this gap is added to the existing layer.
 
-    2 part of the new layer lies within an existing layer, bottom is never below: layer
-    is added above the existing layer, and the top of existing layer is lowered.
+    2 part of the new layer lies within an existing layer, but the bottom of the new
+    layer is never below the bottom of the existing layer: layer is added above the
+    existing layer, and the top of existing layer is lowered.
 
     3 there are locations where the new layer is above the bottom of the existing layer,
     but below the top of the existing layer. The new layer splits the existing layer
@@ -1457,9 +1454,6 @@ def insert_layer(ds, name, top, bot, kh=None, kv=None, copy=True):
     todo = ~(np.isnan(top.data) | np.isnan(bot.data)) & ((top - bot).data > 0)
     if not todo.any():
         logger.warning(f"Thickness of new layer {name} is never larger than 0")
-    if copy:
-        # make a copy, so we are sure we do not alter the original DataSet
-        ds = ds.copy(deep=True)
     isplit = None
     for layer in ds.layer.data:
         if not todo.any():
--- a/nlmod/dims/layers.py
+++ b/nlmod/dims/layers.py
@@ -684,7 +684,7 @@ def add_kh_kv_from_ml_layer_to_ds(
     return ds
 
 
-def set_model_top(ds, top, min_thickness=0.0):
+def set_model_top(ds, top, min_thickness=0.0, copy=True):
     """Set the model top, changing layer bottoms when necessary as well.
 
     If the new top is higher than the previous top, the extra thickness is added to the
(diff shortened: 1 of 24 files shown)


nlmod/read/regis.py
```diff
@@ -228,13 +243,21 @@ def add_geotop_to_regis_layers(
...
             gt = geotop.add_kh_and_kv(gt, geotop_k, anisotropy=anisotropy)
 
+    # copy the regis-dataset, before altering it
+    rg = rg.copy(deep=True)
+    if "layer" in rg["top"].dims:
+        msg = "Top in rg has a layer dimension. add_geotop_to_regis_layers will remove the layer dimension from top in rg."
+        logger.warning(msg)
+    else:
+        # temporarily add layer dimension to top in rg
+        rg["top"] = rg["botm"] + calculate_thickness(rg)
```
dbrakenhoff, line 253: instead of replacing the 2d top we want, can we just introduce a temporary variable?

tests/test_009_layers.py
```diff
@@ -108,21 +107,28 @@ def test_set_layer_botm(plot=False):
...
     # just replace the 2nd layer by a new insertion
     layer = ds1.layer.data[1]
     new_layer = "test"
+    ds1_top3d = ds1["botm"] + nlmod.layers.calculate_thickness(ds1)
     ds2 = nlmod.layers.insert_layer(
-        ds1, "test", ds1["top"].loc[layer], ds1["botm"].loc[layer]
+        ds1, "test", ds1_top3d.loc[layer], ds1["botm"].loc[layer]
     )
+
+    # make sure the total thickness of the two models is equal (does not assert to True yet)
```
dbrakenhoff, line 121: Why is that... ?

tests/test_009_layers.py
```diff
@@ -108,21 +107,28 @@ def test_set_layer_botm(plot=False):
...
     assert nlmod.layers.calculate_thickness(ds2).loc[layer].sum() == 0
-    mask = ~np.isnan(ds1["top"].loc[layer])
-    assert (
-        ds2["top"].loc[new_layer].data[mask] == ds1["top"].loc[layer].data[mask]
-    ).all()
-    assert (
-        ds2["botm"].loc[new_layer].data[mask] == ds1["botm"].loc[layer].data[mask]
-    ).all()
+
+    # make sure the top of the new layer is euqal to the top of the original layer
```
dbrakenhoff, line 129: comment typo, euqal

dbrakenhoff (commented):
Looks good! 

Two comments to discuss:
* one about the naming of the fill_top_and_bottom() method and what it does. 
* can we think of names for the two layer model data models that we can reference by name, and explain somewhere in the documentation. e.g.  "full-3D layer model" and "Modflow layer model". Maybe we can also use those names in our function names `convert_to_modflow_layermodel` and `convert_to_3d_layermodel()`. And in that explanation we should show how inactive zones can be derived from the thickness attribute. Come to think of it we should maybe add a method, `get_active_domain(ds, layer=None)`.

Anyway, those were my thoughts.

nlmod/dims/layers.py
```diff
@@ -1025,27 +1025,28 @@ def fill_top_and_bottom(ds, drop_layer_dim_from_top=True):
...
+
     if "layer" in ds["top"].dims:
         # remove nans from top by setting it equal to botm
         # which sets the layer thickness to 0
-        top = ds["top"].data
-        mask = np.isnan(top)
-        top[mask] = botm[mask]
+        ds["top"] = ds["top"].where(~ds["top"].isnull(), ds["botm"])
+
+        if drop_layer_dim_from_top:
```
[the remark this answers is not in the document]
dbrakenhoff: Maybe use the layer model names we decide on instead of 2d_top and 3d_top.

tests/test_009_layers.py
rubencalje: No idea yet... It is a difference of 0.125 if I sum the total thickness. I have no idea where this comes from.

dbrakenhoff (commented):
Add
- `convert_to_modflow_layermodel()`,  alias for remove_layer_dim_from_top + remove NaNs
- `convert_to_regis_layermodel()`, alias for add_layer_dim_to_top + add NaNs

dbrakenhoff (approved):

PR #290, opened 2023-11-14
Improvements gdf to da

bdestombe:
Improvements to intersecting shapes to grid functions.

The buffer argument I would like to use for identifying transition zones between two layer_models.

Introduced a new function that counts in how many shapes a cell is present. Basically all the code was already there

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ transport models at different spatial and temporal scales to answer specific
 geohydrological questions. Scripting these steps, from downloading data to building
 groundwater models, makes models more reproducible and transparent.
 
-The functions in nlmod have four main objectives:
+The functions in `nlmod` have four main objectives:
 
 1. Create and adapt the temporal and spatial discretization of a MODFLOW model using an xarray Dataset (`nlmod.dims`).
 2. Download and read data from external sources, project this data on the modelgrid and add this data to an xarray Dataset (`nlmod.read`).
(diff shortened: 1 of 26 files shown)


dbrakenhoff:
> Is there a flopy function to leave out parts of a vertex modelgrid? I would like to try it to see if it would speed up gdf_to_da

I'm not aware of any such functionality, though it should be fairly easy to use gridintersect to create a subset of geometries based on an intersection result. Recreating a model grid from that object would be a bit difficult though. But since you need to build the intersection geometries for that step anyway, I doubt it will speed things up much leaving them out of the next intersection calculations? The shapely intersection routines should be able to query the relevant parts of the grid relatively quickly, I think?

bdestombe:
The PR currently has all the improvents I was looking for. According to codacy there is a space too many. However, the tests stopped passing after PR #803, so passing that that seems out of the scope for this PR.

In a future PR it would be interesting to also vectorize the "aggregate values based on nearest to cell node"-function, and other aggregation functions.

rubencalje:
Nice. I have fixed the failing test in pull request #289, so I think this is ok to merge. We should add some test for gdf_to_da, gdf_to_bool_da and the other methods in grid.py...

bdestombe:
> > Is there a flopy function to leave out parts of a vertex modelgrid? I would like to try it to see if it would speed up gdf_to_da
> 
> I'm not aware of any such functionality, though it should be fairly easy to use gridintersect to create a subset of geometries based on an intersection result. Recreating a model grid from that object would be a bit difficult though. But since you need to build the intersection geometries for that step anyway, I doubt it will speed things up much leaving them out of the next intersection calculations? The shapely intersection routines should be able to query the relevant parts of the grid relatively quickly, I think?

Perfect, thank you so much for your response. I'll leave it at that then.

rubencalje (approved):

PR #295, opened 2023-11-22
Split layer ds minor changes

bdestombe:
Merge after #294

### DIFF
--- a/nlmod/dims/base.py
+++ b/nlmod/dims/base.py
@@ -77,6 +77,7 @@ def to_model_ds(
     angrot=0.0,
     drop_attributes=True,
     transport=False,
+    remove_nan_layers=True,
 ):
     """Transform an input dataset to a groundwater model dataset.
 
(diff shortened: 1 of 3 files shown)


nlmod/dims/base.py
```diff
@@ -125,6 +126,9 @@ def to_model_ds(
     transport : bool, optional
         flag indicating whether dataset includes data for a groundwater
         transport model (GWT). Default is False, no transport.
+    remove_nan_layers : bool, optional
```
dbrakenhoff, line 129: we should probably rename this to 'remove_inactive_layers' or 'remove_zero_thickness_layers'.

With the move to internally always assume modflow-style layer models (with one top and N bottoms), renaming this kwarg would make it more clear what we're doing. 

We do need to maintain some backward compatibility though, so we should add the new name and throw a DeprecationWarning for the old one. Up to you whether you open a new issue for this or add it as part of this PR :).

dbrakenhoff (approved):
One comment, but can also be solved in another issue. Up to you.

nlmod/dims/base.py
bdestombe: #300

PR #297, opened 2023-11-24
Improve output

rubencalje:
Add add_nans-argument to nlmod.gwf.output.get_heads_da
todo: add tests for output

### DIFF
--- a/nlmod/mfoutput/mfoutput.py
+++ b/nlmod/mfoutput/mfoutput.py
@@ -77,7 +77,7 @@ def _get_time_index(fobj, ds=None, gwf_or_gwt=None):
     return tindex
 
 
-def _create_da(arr, modelgrid, times):
+def _create_da(arr, modelgrid, times, add_nans=True):
     """Create data array based on array, modelgrid, and time array.
 
     Parameters
@@ -88,6 +88,8 @@ def _create_da(arr, modelgrid, times):
         flopy modelgrid object
     times : list or array
         list or array containing times as floats (usually in days)
+    add_nans : bool, optional
+        If True, sets dry/no-flow cells to nan. The default is True.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


nlmod/mfoutput/mfoutput.py
```diff
@@ -88,6 +88,12 @@ def _create_da(arr, modelgrid, times):
         flopy modelgrid object
     times : list or array
         list or array containing times as floats (usually in days)
+    add_nans : bool, optional
```
dbrakenhoff, line 91: Nothing major, but maybe `dry_nans = True` is a bit clearer, instead of add_nans.

tests/test_002_regis_geotop.py
```diff
@@ -48,3 +48,18 @@ def test_get_regis_geotop_keep_all_layers(
...
+        [118200, 118300, 439800, 439900], probabilities=True
+    )
+
+    df = nlmod.read.geotop.get_lithok_props()
+    # stochastic = None/False is allready tested in methods above
+    gt = nlmod.read.geotop.add_kh_and_kv(gt, df, stochastic=True)
+
+    df = nlmod.read.geotop.get_kh_kv_table()
+    gt = nlmod.read.geotop.add_kh_and_kv(gt, df)
+    # again, but using the stochastic method
```
dbrakenhoff, line 64: seems like stochastic is already tested in line before? Maybe modify comment to indicate why second test is different from first?

dbrakenhoff (approved):
2 very minor comments.

tests/test_002_regis_geotop.py
rubencalje: Added more comments

nlmod/dims/grid.py
```diff
@@ -113,6 +113,28 @@ def xy_to_icell2d(xy, ds):
     return icell2d
 
 
+def xy_to_row_col(xy, ds):
```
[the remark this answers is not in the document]
rubencalje: Added check, also in xy_to_icell2d. But I think we should improve these methods, as they will give a wrong result in variable gridsizes, where the closest centre of the cell in not equal to the cell the point is in.

PR #304, opened 2023-12-06
gdf_to_da should return array with fill_value if shape is outside grid

bdestombe:


dbrakenhoff (approved):

dbrakenhoff:
Test fails unrelated.

PR #307, opened 2023-12-13
Fix remove layer

rubencalje:
When the first layer is removed, the top needs to be set on the botm of this layer, not on the botm of the second layer.

### DIFF
--- a/nlmod/dims/layers.py
+++ b/nlmod/dims/layers.py
@@ -239,9 +239,9 @@ def split_layers_ds(
             split_dict[lay0] = [1 / split_dict[lay0]] * split_dict[lay0]
         elif hasattr(split_dict[lay0], "__iter__"):
             # make sure the fractions add up to 1
-            assert np.isclose(np.sum(split_dict[lay0]), 1), (
-                f"Fractions for splitting layer '{lay0}' do not add up to 1."
-            )
+            assert np.isclose(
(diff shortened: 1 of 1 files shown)


rubencalje:
Tests fail unrelated to thsi pr

nlmod/dims/layers.py
```diff
@@ -1815,8 +1815,8 @@ def remove_layer(ds, layer):
         ds = ds.copy(deep=True)
         index = layers.index(layer)
         if index == 0:
-            # lower the top to the second layer
-            ds["top"] = ds["botm"].loc[layers[1]]
+            # lower the top to the botm of the first layer
+            ds["top"] = ds["botm"].loc[layers[0]]
```
OnnoEbbens, line 1819: I think the docstring needs to be changed. It says now: "Removes a layer from a Dataset, **without changing elevations of other layers**". The last part is (and was already) incorrect.

OnnoEbbens (changes requested):

nlmod/dims/layers.py
rubencalje: I think the docstring is (still) correct. The top and botm of other layers are not changed. Only when the top layer is removed, the top of the new top layer (that used to be the second layer) is set to the variable `top`. The top of the second layer is not changed, only the way it is represented in ds changes.

PR #309, opened 2023-12-21
Release 0.7.2

OnnoEbbens:
Lot's of small and somewhat bigger fixes have been made so it's time for an end of the year release.


rubencalje:
In the documentation-website, the following warning is shown at many examples:

`WARNING:py.warnings._showwarnmsg:/home/docs/checkouts/readthedocs.org/user_builds/nlmod/envs/latest/lib/python3.10/site-packages/xarray/core/utils.py:494: FutureWarning: The return type of `Dataset.dims` will be changed to return a set of dimension names in future, in order to be more consistent with `DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.
  warnings.warn(`

If possible, we need to solve this before the next version I think.

PR #314, opened 2024-01-23
Update webservices.py

bdestombe:
Two issues are solved:
- If values for a large area are prompted, the query url gets very long and is not accepted by HHNK polderpeil server. Is now queried in chunks
- An error in the database of HHNK resulted in duplicate peil definitions. Only use the unambigious entries and leave a note in the logs.

### DIFF
--- a/nlmod/read/webservices.py
+++ b/nlmod/read/webservices.py
@@ -150,22 +150,53 @@ def arcrest(
         else:
             gdf = gpd.GeoDataFrame.from_features(features, crs=sr)
             if table is not None:
-                url_query = f"{url}/{table.pop('id')}/query"
-                pgbids = ",".join([str(v) for v in gdf["OBJECTID"].values])
-                params["where"] = f"PEILGEBIEDVIGERENDID IN ({pgbids})"
                 params["f"] = "json"
-                data = _get_data(url_query, params, timeout=timeout)
+                url_query = f"{url}/{table.pop('id')}/query"
+
+                # loop over chunks of 100 pgbids. Long where clauses can cause
+                # the request to fail. 1300 pgbids fails but 130 works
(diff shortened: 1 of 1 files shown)


rubencalje:
This PR adds waterboard-specific info (eg "PEILGEBIEDVIGERENDID" to webservices.py, just like previous commmits 
1e66726fc1b05f2d7994cd6685adea7a16c99014 and 1e66726fc1b05f2d7994cd6685adea7a16c99014. I think we should separate the waterboard-specific information from the reading of the data.

bdestombe:
Hi Ruben,
Thanks for your response. This PR indeed builds on top of the commits of Onno. I am not entirely sure how to proceed here, as I am not that familiar with the various gis servers.

rubencalje:
OK, sorry for the blunt comment then. Let's continue the way you propose, and think of an approach to seperate the waterboard-specific information in the future. @OnnoEbbens has some weeks off, so we can better wait unitil he is back with large changes. We can merge this PR I think.

OnnoEbbens:
Ah I understand. I should have put the logic for waterboards in waterboards.py and not in webservices.py. I don't remember why I did that. I can fix this, but unfortunately not anytime soon. For now we can merge this as a quick fix and leave an issue open for me to fix this correctly later.

OnnoEbbens (approved):

bdestombe:
Oh no, there were no harsh feelings or what so ever. I just didn't have an
idea ready to solve the problem you brought up.
Do the commits of Onno and me break the function when used for other
waterboards?

OnnoEbbens:
No harsh feelings from me either :)

I think the arcrest function is only used in waterboards.py so it should not break anything. @rubencalje: please correct me if I am wrong

PR #324, opened 2024-02-12
Zero layer thickness layers are considered active due to float rounding

bdestombe:
Subtracting two botms to calculate the thickness results is some layers with a thickness of ~1e-9. This rounding later results in a problem because this positive thickness leads to an idomain of 1.

This can either be solved in `nlmod.layers.calculating_thickness()` or in `nlmod.layers.get_idomain()`. What do you prefer?

### DIFF
--- a/nlmod/dims/layers.py
+++ b/nlmod/dims/layers.py
@@ -1314,6 +1314,8 @@ def get_idomain(ds):
     idomain.attrs.clear()
     # set idomain of cells  with a positive thickness to 1
     thickness = calculate_thickness(ds)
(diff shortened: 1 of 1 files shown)


bdestombe:
> Maybe both? Calculating a thickness is just a subtraction operation, so theoretically if small values appear in that calculation it isn't necessarily a problem. It does become a problem when we assume it contains zeros for creating a mask or whatever. So we definitely should include a minimum thickness tolerance in `get_idomain()`. But it would also be nice to just get zeros where we expect them in `calculate_thickness()`.

I made the change. Can I merge it like this?

dbrakenhoff (approved):

PR #325, opened 2024-02-15
Improve grid rotation

rubencalje:


nlmod/dims/base.py
OnnoEbbens, line 1: the docstrings of `get_ds` and `_set_angrot_attributes` mention for "xorigin" and "yorigin": "Only used when angrot is not 0." I think this is incorrect and they are also used if angrot is 0. It would be nice to have a bit more information in the docstring about what happens to the extent when you use `xorigin`, `yorigin` and `angrot`. What coordinates are there in the dataset model coördinates, real-world-coordinates.

nlmod/dims/grid.py
```diff
@@ -1536,7 +1542,7 @@ def gdf_to_count_da(gdf, ds, ix=None, buffer=0.0, **kwargs):
 
     # build list of gridcells
     if ix is None:
-        modelgrid = modelgrid_from_ds(ds)
+        modelgrid = modelgrid_from_ds(ds, rotated=False)
```
OnnoEbbens, line 1545: Just checking if I understand correctly: You convert the gdf to modelcoordinates if you have a rotated grid. Then you use `rotated=False` for the modelgrid because you need a modelgrid with modelcoordinates (and not real world coördinates).

nlmod/dims/resample.py
```diff
@@ -139,11 +139,11 @@ def ds_to_structured_grid(
     if delc is None:
         delc = delr
```
OnnoEbbens, line 141: see previous comment about xorigin and yorigin in docstring.

nlmod/dims/resample.py
```diff
@@ -431,7 +436,9 @@ def dim_to_regular_dim(da, dims, z):
         coords = dict(da.coords)
         coords["x"] = ds.x
         coords["y"] = ds.y
-        coords.pop("icell2d")
+        for key in list(coords):
```
OnnoEbbens, line 439: Why is this necessary? I think 'icell2d' is always in the coordinates of da because at the beginning of the function this checked.

tests/test_026_grid.py
OnnoEbbens, line 1: Looks neat!

nlmod/dims/base.py
rubencalje: There is a notebook for that: https://nlmod.readthedocs.io/en/latest/examples/11_grid_rotation.html

nlmod/dims/grid.py
rubencalje: Yes

nlmod/dims/resample.py
rubencalje: We need to delete all coordinates that have icell2d as a dimension, like 'xc', 'yc'. The coordinate 'icell2d' also has 'icell2d' as a dimension, so it is also removed. So no need to specifically delete it, this saves a line.

nlmod/dims/base.py
rubencalje: You are right, xorigin and yorigin are used when angrot == 0. Will change the docstrings.

rubencalje:
I implemented the requested changes, by improving the docstring of xorigin, yorigin and angrot.

nlmod/dims/resample.py
OnnoEbbens: Ah that makes sense, thanks for explaining

OnnoEbbens (approved):

PR #345, opened 2024-05-17
Cache: added shorthand for model ds attributes

bdestombe:

PR #347, opened 2024-05-22
Added get_exe_path with more features

bdestombe:
Use previously installed anywhere. If doesn't exist, download it.

### DIFF
--- a/nlmod/util.py
+++ b/nlmod/util.py
@@ -7,7 +7,6 @@ from pathlib import Path
 import warnings
 from typing import Dict, Optional
 
-import flopy
 from flopy.utils import get_modflow
 from flopy.utils.get_modflow import flopy_appdata_path, get_release
 import geopandas as gpd
@@ -19,6 +18,7 @@ from shapely.geometry import box
 logger = logging.getLogger(__name__)
 
 nlmod_bindir = Path(__file__).parent / "bin"
+
 
 class LayerError(Exception):
     """Generic error when modifying layers."""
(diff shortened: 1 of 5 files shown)


bdestombe:
As I am switching a lot between different Python environments and recreating them, the implemented features greatly reduce the number of having to download the executables.

bdestombe:
The tests failing are not related to this PR. They were also failing in the previous commit: #345

nlmod/util.py
```diff
@@ -89,29 +94,276 @@ def get_model_dirs(model_ws):
...
+    return exe_full_path
+
+
+def get_bin_directory(
+    bindir=None,
+    exe_name="mf6",
+    download_if_not_found=True,
+    version_tag="latest",
+    repo="executables",
+    disable_version_check=False,
```
dbrakenhoff, line 168: Nicer to have a single negative? `enable_version_check=True`?

nlmod/util.py
```diff
@@ -89,29 +94,276 @@ def get_model_dirs(model_ws):
...
+    if sys.platform.startswith("win") and not exe_name.endswith(".exe"):
+        exe_name += ".exe"
+
+    # If bindir is provided
+    if bindir is not None and not disable_version_check:
+        msg = "Incompatible arguments. If bindir is provided, disable_version_check should be True."
+        raise ValueError(msg)
+
+    use_bindir = (
+        bindir is not None and exe_name is not None and Path(bindir / exe_name).exists()
```
dbrakenhoff, line 224: Don't think you need to wrap `bindir/ exe_name` in another Path call, it is already a pathlib.Path I think?

nlmod/util.py
```diff
@@ -89,29 +94,276 @@ def get_model_dirs(model_ws):
...
+    if not flopy_metadata_fp.exists():
+        return []
+
+    meta_raw = flopy_metadata_fp.read_text()
+
+    # Remove trailing characters that are not part of the JSON.
+    while meta_raw[-3:] != "}\n]":
+        meta_raw = meta_raw[:-1]
+
+    # get metadata of most all installations
```
dbrakenhoff, line 311: "most all" ?

nlmod/util.py
dbrakenhoff: In case this is a flopy kwarg, it's fine to use that so we match flopy's logic.

bdestombe: Indeed! Great suggestion!

PR #349, opened 2024-05-27
Github token needed for CI when downloading executables

bdestombe:
Also required by flopy tests: https://github.com/modflowpy/flopy/blob/388ac8d484b08ae26a57793759bae22dd0db21fb/.github/workflows/mf6.yml#L81

edit: suppressing checking version of exe during tests instead of adding github token

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,8 @@ jobs:
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=80 --statistics
 
       - name: Download executables needed for tests
+        env:
+          GITHUB_TOKEN: ${{ github.token }}  # Needed for downloading executables
         shell: bash -l {0}
         run: |
           python -c "import nlmod; nlmod.util.download_mfbinaries()"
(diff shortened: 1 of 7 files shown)


bdestombe:
The reason it is suddenly exceeding the interaction rate github allows without github token is because now the version of modflow is often checked. This PR includes a way to suppress the version checking by setting an environment variable, which is now used in the github CI.

OnnoEbbens:
I am back from the holidays and saw many improvements so I might not be completely up to date. Do I understand correctly that every time you call `get_exe_path()` using the default options a check is done to see if your local modflow executable is the latest version? And then these checks are done so many times that interaction rate is exceeded?

If that is correct than I am wondering if we really want to do this check every time. I personally almost never want to download the newest executable when I call the `get_exe_path` function. I just want to use the one that I've used before. So instead of setting an environment variable can we not make `enable_version_check=False` by default?

bdestombe:
Hi Onno,
Thanks for having a look :)

So I looked it up. You can do 60 requests per hour to the github website without account/token. That means 1 time downloading and 59 times converting "latest" to "18.0". And is called with every to_model_ds(), so this is indeed not feasible. I'll change it.

PR #352, opened 2024-06-17
Fix regis URL

rubencalje:
This PR simplifies the REGIS- and GeoTOP-urls, to the openda[-server named on https://www.dinoloket.nl/modelbestanden-aanvragen/netcdf. It also sets the version of the netcdf4 package to below 1.70, because of issue #353.

### DIFF
--- a/nlmod/read/geotop.py
+++ b/nlmod/read/geotop.py
@@ -13,7 +13,7 @@ from ..util import MissingValueError
 
 logger = logging.getLogger(__name__)
 
-GEOTOP_URL = r"http://www.dinodata.nl/opendap/GeoTOP/geotop.nc"
+GEOTOP_URL = "https://dinodata.nl/opendap/GeoTOP/geotop.nc"
 
 
 def get_lithok_props(rgb_colors=True):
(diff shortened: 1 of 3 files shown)


bdestombe:
Hi Ruben,
Why did you limit the netCDF4 versions only in the CI and the creation of the documentation, and not in the main dependency list of nlmod (https://github.com/gwmod/nlmod/blob/b7fb73f6a9b2cf79efcdc74e508e43d2a44a7228/pyproject.toml#L26 )?

rubencalje:
> Hi Ruben,
> Why did you limit the netCDF4 versions only in the CI and the creation of the documentation, and not in the main dependency list of nlmod (https://github.com/gwmod/nlmod/blob/b7fb73f6a9b2cf79efcdc74e508e43d2a44a7228/pyproject.toml#L26 )?

Because I could not reprise the error on my Windows pc, so I thought maybe it is only a linux-problem. And I did not want to restrict package versions too much.

bdestombe:
Ah but what about Davíd and the few other crazy Linux users? ;)

In my case, I use nlmod as a dependency and the CI of my scripts are failing because of it. To me it would have made sense if it would have been added to the main dependecy list of nlmod.

PR #354, opened 2024-06-24
Patch auto-download executables

### LINKED ISSUE #348: Remove explicit download of executables command from examples
Shall we remove the explicit command at the beginning of the notebooks to download the executables and instead rely on auto-download mf-executables in the new `get_exe_path()`?

bdestombe:
- download_mfbinaries() should not retreive any None values for version_tags. Which was the case if: no mf6 was ever downloaded on your system and no version_tag was specified in get_bin_directory() or to_model_ds() .
- Remove explicit download commands and rely on auto download of executables. Closes #348

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,11 +39,6 @@ jobs:
           # exit-zero treats all errors as warnings.
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=80 --statistics
 
-      - name: Download executables needed for tests
-        shell: bash -l {0}
-        run: |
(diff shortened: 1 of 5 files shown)


bdestombe:
Uhhg. They have built a nasty conditional statement that detects testing environment and in case of testing don't write the metadata. The written metadata is needed later for locating installs. https://github.com/modflowpy/flopy/blob/0748dcb9e4641b5ad9616af115dd3be906f98f50/flopy/utils/get_modflow.py#L623

A bit like the Volkswagen/audi/diesel gate that detected when the cars were tested for emissions. During testing conditions they would emit less..

dbrakenhoff (approved):

PR #357, opened 2024-06-27
add modelextent plot

dbrakenhoff:
This PR is originally made by @dbrakenhoff to add a modelextent-plot.

It has been misused by me (@rubencalje) to replace some of the methods from the resample-module to the grid-module. For example the `nlmod.resample.get_extent(ds)` now is `nlmod.grid.get_extent(ds)`, which I think is more logical. The original methods still function, and the user will see a warning that the old methods are deprecated.

### DIFF
--- a/nlmod/dims/resample.py
+++ b/nlmod/dims/resample.py
@@ -615,7 +615,7 @@ def affine_transform_gdf(gdf, affine):
 def get_extent(ds, rotated=True):
     """Get the model extent, corrected for angrot if necessary."""
     logger.warning(
-        "nlmod.resample.get_extent is deprecated. " "Use nlmod.grid.get_extent instead"
+        "nlmod.resample.get_extent is deprecated. Use nlmod.grid.get_extent instead"
     )
     from .grid import get_extent
 
@@ -647,8 +647,8 @@ def get_affine_world_to_mod(ds):
 def get_affine(ds, sx=None, sy=None):
     """Get the affine-transformation, from pixel to real-world coordinates."""
     logger.warning(
-        "nlmod.resample.get_affine is deprecated. " "Use nlmod.grid.get_affine instead"
+        "nlmod.resample.get_affine is deprecated. Use nlmod.grid.get_affine instead"
     )
     from .grid import get_affine
 
-    return get_affine(ds)
+    return get_affine(ds, sx=sx, sy=sy)
--- a/nlmod/dims/resample.py
+++ b/nlmod/dims/resample.py
@@ -6,10 +6,7 @@ import rasterio
 import xarray as xr
 from scipy.interpolate import griddata
 from scipy.spatial import cKDTree
-
-
 from ..util import get_da_from_da_ds
-
 
 logger = logging.getLogger(__name__)
 
--- a/nlmod/dims/resample.py
+++ b/nlmod/dims/resample.py
@@ -4,13 +4,12 @@ import numbers
 import numpy as np
 import rasterio
 import xarray as xr
-from affine import Affine
 from scipy.interpolate import griddata
 from scipy.spatial import cKDTree
-from shapely.affinity import affine_transform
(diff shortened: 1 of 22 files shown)


dbrakenhoff:
Test fails unrelated.

rubencalje:
> Nice addition. There are already two functions in nlmod.util to create a polygon from the extent `polygon_from_extent` and to create a geodataframe from an extent `gdf_from_extent`. I think these functions are nice to keep and the plot function can call them. Only thing is that we should move them somewhere else because 'util' is not the best place. Probably the `resample` module is a good place because `get_extent` is also in there.

It is even worse... We also have `nlmod.resample.get_extent_polygon`. This method also takes a rotated grid into account.

rubencalje:
> > Nice addition. There are already two functions in nlmod.util to create a polygon from the extent `polygon_from_extent` and to create a geodataframe from an extent `gdf_from_extent`. I think these functions are nice to keep and the plot function can call them. Only thing is that we should move them somewhere else because 'util' is not the best place. Probably the `resample` module is a good place because `get_extent` is also in there.
> 
> It is even worse... We also have `nlmod.resample.get_extent_polygon`. This method also takes a rotated grid into account. Maybe we can merge some of these methods?

I moved some of the methods from the resample-module to the grid-module (also see the starting-message of this PR). I also moved `extent_to_polygon` from the resample-module to the util-module, where it replaces `polygon_from_extent`, and renamed `gdf_from_extent` to `extent_to_gdf` in the util-module. These methods in the util-module can also be used without a grid, and therefore the util-module is a good place for these methods.

All the old methods continue to work for now, and the user will get to see a deprecation-message.

nlmod/dims/resample.py
```diff
@@ -573,6 +518,8 @@ def structured_da_to_ds(da, ds, method="average", nodata=np.nan):
             raise ValueError(
                 "No extent or delr and delc in ds. Cannot determine affine."
             )
+        from .grid import get_affine
```
OnnoEbbens, line 521: why not a top level import? It is now a separate import for structured and vertex.

nlmod/plot/plot.py
```diff
@@ -48,6 +52,40 @@ def modelgrid(ds, ax=None, **kwargs):
...
+    """Plot model extent.
+
+    Parameters
+    ----------
+    ds : xarray.Dataset
+        The dataset containing the data.
+    ax : matplotlib.axes.Axes, optional
+        The axes object to plot on. If not provided, a new figure and axes will be
+        created.
+    **kwargs
```
OnnoEbbens, line 65: Can you add `dx`  and `rotated` parameters to the docstring. Should we make `rotated` False by default?

OnnoEbbens (changes requested):
Compliments for all the proper deprecations. I just have some minor comments.

nlmod/plot/plot.py
dbrakenhoff: Added, I kept the rotated=True default for now. Deferring that decision to @rubencalje .

nlmod/dims/resample.py
rubencalje: Because the deprecated version of get_affine is also in resample.py. We do not want the get_affine from resample, but from grid. Not very pretty, but for now it works, and the user gets to see the deprecation message when he uses nlmod.resample.get_affine.

nlmod/plot/plot.py
rubencalje: I think @OnnoEbbens has a good point, as we have specified the following in the grid-rotation-notebook (https://nlmod.readthedocs.io/en/latest/examples/11_grid_rotation.html):
`the plot-methods in nlmod plot the grid in model-coordinates by default (can be overridden by the setting the parameter ‘rotated’ to True)`

Since then, new plot-methods have adapted a default of rotated=True. I will change these defaults to False

Issue #356, opened 2024-06-27
Improve CI


dbrakenhoff:
- Remove flake8 check (unused anyway).
- Move CI to Python 3.11?
- Up versions of actions used in CI for checkout and setup-python actions. 
- Remove KNMI Data Platform tests, they seem to be failing quite often?
- Apply auto fixes suggested by Ruff, e.g. for numpy 2.0.
```
Found 440 errors. [*] 117 fixable with the `--fix` option
```
- Start gradually fixing all other ruff issues (mostly simple things like missing docstrings). Maybe GH Copilot can help. 
- ...

PR #358, opened 2024-06-27
Improve CI and some code cleanup

dbrakenhoff:
See #356.


bdestombe:
I am a bit uncomfortable with skipping the KNMI test, but I am also not fully understanding why it fails.
File not found. Should we introduce logic that solves for the missing data? Or instead of skipping the KNMI test, should we alter the test as the test problem as it is posed now should fail? Or is the user never going to end up in a situation that this error would be relevant?

dbrakenhoff:
The test skip is meant to be temporary. When @martinvonk is back from his holiday, I was planning to let him look into it. Meanwhile we can release the next version with nlmod without being bothered by these failing tests. 

I thought the KNMI Data Platform functionality wasn't that extensively used, so it would be okay to fix later.

OnnoEbbens:
The test that was skipped is for the KNMI data platform to obtain temporally and spatially distributed meteodata. In the examples we don't use the data from the KNMI data platform. We use the KNMI data obtained via hydropandas. If I remember correctly it is not very useful to use the data from the KNMI data platform yet as there are some weird interpolation results which might mess up the model. Also if you use temporally and spatially distributed meteodata in the model it usually becomes very slow. So we don't have a very good use case for this data yet.

bdestombe:
Ah, I wasn't aware that this was an alternative route for obtaining KNMI data. Thanks both of you for clearing this up.

Issue #367, opened 2024-08-20
Add support for Algemeen Hoogtebestand Nederland versie 5 (AHN5)


martinvonk:
AHN5 is released in 2023 for northwest part of the Netherlands and more recently for the western part of the Netherlands. The southern part will be available in 2024, and the northeastern part in 2025: https://www.ahn.nl/tweede-deel-ahn5-beschikbaar
![map](https://cuatro.sim-cdn.nl/ahn/uploads/styles/wysiwyg_medium/media/vliegkaart_2.jpg?cb=RQ7dfE8Z)

Maybe we can add support for the already available tiles in the AHN5:
DTM_WMS: https://api.ellipsis-drive.com/v3/ogc/wms/51e11c02-1065-463d-a0b0-263d80293b16
DTM_WMTS: https://api.ellipsis-drive.com/v3/ogc/wmts/51e11c02-1065-463d-a0b0-263d80293b16
DTM_WCS: https://api.ellipsis-drive.com/v3/ogc/wcs/51e11c02-1065-463d-a0b0-263d80293b16

DSM_WMS: https://api.ellipsis-drive.com/v3/ogc/wms/a4a8a27b-e36e-4dd5-a75b-f7b6c18d33ec
DSM_WMTS: https://api.ellipsis-drive.com/v3/ogc/wmts/a4a8a27b-e36e-4dd5-a75b-f7b6c18d33ec
DSM_WCS: https://api.ellipsis-drive.com/v3/ogc/wcs/a4a8a27b-e36e-4dd5-a75b-f7b6c18d33ec

Tiles example for 38DN1:
LAZ: https://ns_hwh.fundaments.nl/hwh-ahn/AHN5/01_LAZ/2023_C_38DN1.LAZ
DTM 0.5m: https://ns_hwh.fundaments.nl/hwh-ahn/AHN5/02a_DTM_50cm/2023_M_38DN1.TIF
DTM 5.0m: https://ns_hwh.fundaments.nl/hwh-ahn/AHN5/02b_DTM_5m/2023_M5_38DN1.TIF
DSM 0.5m: https://ns_hwh.fundaments.nl/hwh-ahn/AHN5/03a_DSM_50cm/2023_R_38DN1.TIF
DSM 5.0m: https://ns_hwh.fundaments.nl/hwh-ahn/AHN5/03b_DSM_5m/2023_R5_38DN1.TIF

Nice gimmick:
[...]

Issue #369, opened 2024-09-17
Download ahn produces only zeros with rasterio version 1.3.11


rubencalje:
Showing warning:

> WARNING:py.warnings._showwarnmsg:C:\Users\Ruben\miniconda3\envs\artesia\lib\site-packages\rasterio\merge.py:321: UserWarning: Ignoring nodata value. The nodata value, 3.4028234663852886e+38, cannot safely be represented in the chosen data type, float32. Consider overriding it using the --nodata option for better results. Falling back to first source's nodata value.

dbrakenhoff:
Fixed by #370

bdestombe:
The ellipsis server used in the new method introduced in #370 seems to lack a few tiles. Some urls are nan's leading to errors further down the line.
![image](https://github.com/user-attachments/assets/8957a4a6-8aec-4564-b41e-746fac32d22c)

rubencalje:
Nice catch.

I think it is just the tile in the northsea, that has been renamed between version 2 and 3 of ahn, I guess. See the location of tiles 72Hz2 and 18HZ2 below. I will add a check that ignores nan-values in the tile-names. This is probably also needed for AHN 5 probably, which is not defined everywhere in the Netherlands jet.

![18HZ2](https://github.com/user-attachments/assets/cadc8981-80e5-4334-9dd8-0d0d1944235e)
![72HZ2](https://github.com/user-attachments/assets/ccc96ae5-9c45-41d1-9a12-ac8b49099829)

bdestombe:
#371

bdestombe:
Closed with #371

PR #370, opened 2024-09-18
Update ahn.py

rubencalje:
This PR fixes issues #367 and #369, by adding support to AHN 5, and implementing a different way to retrieve the ahn-data.

This PR changes the way we retrieve the AHN data for all AHN-versions, to use the Ellipsis Drive (described in https://www.ahn.nl/dataroom).

`get_ahn1`, `get_ahn2`, `get_ahn3` and `get_ahn4` have been renamed to `get_ahn1_legacy`, `get_ahn2_legacy`, `get_ahn3_legacy` and `get_ahn4_legacy`, mainly so the user can test if the same output is returned by the new methods: new versions of `get_ahn1`, `get_ahn2`, `get_ahn3`,  `get_ahn4` and `get_ahn5` have been implemented that use the files from Ellipsis Drive. These methods will always return a xarray DataArray, so the use of the `as_data_array`-argument has been deprecated. Because the merging of the different tiles now takes place at the rioxarray-level (and not at the rasterio-level), the bug described in #369 is solved.

The identifier-argument on the ellipsis-drive (which determines the resolution of the downloaded data and if a DTM or DSM is downloaded) is a bit different from the ones we used to use. Some logic has been implemented, so the new methods should also work with the old identifier-names.

### DIFF
--- a/nlmod/read/ahn.py
+++ b/nlmod/read/ahn.py
@@ -40,6 +40,8 @@ def get_ahn(ds=None, identifier="AHN4_5M_M", method="average", extent=None, **kw
             AHN3: 'AHN3_05M_M', 'AHN3_05M_R', 'AHN3_5M_M' or 'AHN3_5M_R'
             AHN4: 'AHN4_05M_M', 'AHN4_05M_R', 'AHN4_5M_M' or 'AHN4_5M_R'
             AHN5: 'AHN5_5M_M', 'AHN5_5M_R', 'AHN5_05M_M' or 'AHN5_05M_R'
+        The identifier determines the resolution (05M for 0.5 m and 5M for 5 m) and the
+        type of height data (M = DTM = surface level, R = DSM = also other features).
         The default is 'AHN4_5M_M'.
     method : str, optional
         Method used to resample ahn to grid of ds. See documentation of
@@ -312,8 +314,8 @@ def _get_tiles_ellipsis(
         raise (HTTPError(f"Request not successful: {r.url}"))
     gdf = gpd.GeoDataFrame.from_features(r.json()["result"]["features"], crs=4326)
     gdf = gdf.to_crs(crs)
-    # remove small digits becuase of tcrs-transformation
-    gdf.geometry = gdf.geometry.apply(_round_coordinates, ndigits=1)
+    # remove small digits becuase of crs-transformation
+    gdf.geometry = gdf.geometry.apply(_round_coordinates, ndigits=0)
 
     gdf = gdf.set_index("AHN")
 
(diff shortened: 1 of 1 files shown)


nlmod/read/ahn.py
```diff
@@ -31,11 +35,12 @@ def get_ahn(ds=None, identifier="AHN4_DTM_5m", method="average", extent=None):
         dataset with the model information.
     identifier : str, optional
         Possible values for the different AHN-versions are (casing is important):
-            AHN1: 'ahn1_5m'
-            AHN2: 'ahn2_05m_i', 'ahn2_05m_n', 'ahn2_05m_r' or 'ahn2_5m'
-            AHN3: 'AHN3_05m_DSM', 'AHN3_05m_DTM', 'AHN3_5m_DSM' or 'AHN3_5m_DTM'
-            AHN4: 'AHN4_DTM_05m', 'AHN4_DTM_5m', 'AHN4_DSM_05m' or 'AHN4_DSM_5m'
-        The default is 'AHN4_DTM_5m'.
+            AHN1: 'AHN1_5M'
```
dbrakenhoff, line 38: Add explanation of M and R :)

nlmod/read/ahn.py
```diff
@@ -303,15 +297,80 @@ def get_ahn4_tiles(extent=None):
...
+    path_id="a9d410ad-a2f6-404c-948a-fdf6b43e77a6",
+    timestamp_id="87a21a71-c39f-4e92-a43b-207bc7dfe714",
+):
+    url = f"{base_url}/path/{path_id}/vector/timestamp/{timestamp_id}/listFeatures"
+
+    r = requests.get(url, timeout=timeout)
+    if not r.ok:
+        raise (HTTPError(f"Request not successful: {r.url}"))
+    gdf = gpd.GeoDataFrame.from_features(r.json()["result"]["features"], crs=4326)
+    gdf = gdf.to_crs(crs)
```
dbrakenhoff, line 314: The transform here is happening using EPSG 28992, not our custom definition. Are we sure things are ending up in the right spot in NL?

nlmod/read/ahn.py
```diff
@@ -331,14 +390,36 @@ def get_ahn1(extent, identifier="ahn1_5m", as_data_array=True):
...
+def get_ahn2(extent, identifier="AHN2_5M_M", as_data_array=None, **kwargs):
     """Download AHN2.
 
     Parameters
     ----------
     extent : list, tuple or np.array
         extent
-    identifier : TYPE, optional
+    identifier : str, optional
+        Possible values are 'AHN2_05M_I', 'AHN2_05M_N', 'AHN2_05M_R' and
```
dbrakenhoff, line 401: explain differences between suffixes?

dbrakenhoff (approved):
One question about the transform from EPSG4326 to EPSG28992, whether that is currently correct, as compared to our background map methods? 

For the rest some guidance on which identifier corresponds to which AHN product would be nice. 

Otherwise looks good, so approving this PR.

PR #371, opened 2024-09-23
Update tile retreival ahn.py

bdestombe:
- On the eclipse server: Skip tile retreival if the link does not exist.
- Legacy server: Rasterio merge returned only zero's since 17 sept. All the merging is now done within xarray.

### DIFF
--- a/nlmod/read/ahn.py
+++ b/nlmod/read/ahn.py
@@ -582,7 +582,7 @@ def _get_ahn_ellipsis(extent, identifier="AHN5_5M_M", **kwargs):
         url = tiles.at[tile, identifier]
         if url == "nan":
             continue
-        elif url.endswith(".zip"):
+        if url.endswith(".zip"):
             path = url.split("/")[-1].replace(".zip", ".TIF")
             if path.lower().endswith(".tif.tif"):
                 path = path[:-4]
--- a/nlmod/read/ahn.py
+++ b/nlmod/read/ahn.py
@@ -580,7 +580,9 @@ def _get_ahn_ellipsis(extent, identifier="AHN5_5M_M", **kwargs):
     das = []
     for tile in tqdm(tiles.index, desc=f"Downloading tiles of {identifier}"):
         url = tiles.at[tile, identifier]
-        if url.endswith(".zip"):
+        if url == "nan":
+            continue
+        elif url.endswith(".zip"):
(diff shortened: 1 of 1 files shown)


rubencalje:
Great.

I would just leave the old method (`_download_and_combine_tiles`) intact, and not change anything. With the proposed changes, and when `as_data_array` is `False`, you return something else as before . This method is just for people who wish to use the exact same output as before, using the legacy-methods. So changing its output-type is kind of contradictory to this. In a few versions, we will remove the legacy-methods and `_download_and_combine_tiles`.

bdestombe:
Well could you maybe confirm? If I use `_download_and_combine_tiles()`, during the merge command of rasterio all values are converted to 0. With the proposed changes the function works like intended.

bdestombe:
Since no error is produced and zeros are falsely returned, I would suggest to either 
- already remove these functions, 
- keep these functions as legacy functions with an additional warning that their outcome is wrong
- keep these suggested changes

rubencalje:
The zeros are caused by rasterio version 1.3.11, and you do get a warning (see issue #369). We changed the normal versions of the methods to use the xarray-merge to solve this bug. However, some people might want to use the old server, or `as_data_array=False`. Therefore, the old methods are still available using the `_legacy`-suffix. However, I do not think we should maintain these methods. This will introduce additional work and errors.

For example. With the proposed changes, the following code
```
import nlmod
extent = [100_000, 110_000, 400_000, 405_000]
nlmod.read.ahn.get_ahn4_legacy(extent)
```

produces

```
<xarray.DataArray 'band_data' (band: 1, y: 0, x: 2000)> Size: 0B
array([], shape=(1, 0, 2000), dtype=float32)
Coordinates:
  * band         (band) int32 4B 1
  * x            (x) float64 16kB 1e+05 1e+05 1e+05 ... 1.1e+05 1.1e+05 1.1e+05
  * y            (y) float64 0B 
    spatial_ref  int32 4B 0
Attributes:
    AREA_OR_POINT:  Area
```

So this is an empty data-array (while the original methods produce the correct data-array, with non-zero data if you do not use rasterio 1.3.11). We can off course fix this issue as well, but this is not the goal of the legacy-methods.

bdestombe:
Thanks Ruben for taking the time. I was falsely under the impression that the all-zeros-bug also appeared with other versions of rasterio. I will revert the code and merge the two remaining lines.

PR #373, opened 2024-09-25
Add zonal_statistics and replace add_info_to_gdf

rubencalje:
And add test

### DIFF
--- a/nlmod/util.py
+++ b/nlmod/util.py
@@ -6,7 +6,9 @@ import sys
 import warnings
 from pathlib import Path
 from typing import Dict, Optional
+from functools import partial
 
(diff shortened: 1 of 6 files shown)


nlmod/util.py
```diff
@@ -1054,3 +1058,191 @@ def _get_value_from_ds_datavar(
             value = value.values
 
     return value
+
+
+def add_info_to_gdf(
```
OnnoEbbens, line 1063: name could be more descriptive, maybe `gdf_intersection_join`. And a docstring woulde be nice :)

OnnoEbbens (changes requested):
Just one comment

PR #376, opened 2024-10-09
add get_isosurface and get_isosurface_1d

dbrakenhoff:
- compute isosurface, e.g. depth at which concentration exceeds a threshold value.


rubencalje (approved):
I approve, but have some questions which could be answered in the docstring of `get_isosurface`:
1 WIll this work for vertex grids as well?
2 Can we use other interpolation techniques than linear?
3 What do the kwargs do in `get_isosurface`?

Issue #254, opened 2023-08-25
Do not add variables/attributes to a Dataset that is used as input


OnnoEbbens:
There are some functions that (unexpectedly) modify the Dataset that is used as input for the function. Mostly because variables or attributes are added to the Dataset. For examples the functions. `ds_to_rch`, `ds_to_evt` and `modelgrid_to_vertex_ds`.

rubencalje:
Another example is `set_ds_time`.

So should we start these methods with a copy of the Dataset?

dbrakenhoff:
If the method has "set_" in the name, I think this is ok, as the function name is pretty clear that it's setting something? But good to review whether this is the desired behavior.

OnnoEbbens:
> If the method has "set_" in the name, I think this is ok, as the function name is pretty clear that it's setting something? But good to review whether this is the desired behavior.

I agree with this as well. So we should not create a copy for function starting with `set_..`. 

Now that I examined `modelgrid_to_vertex_ds` better I think this function has become obsolete since we have the `modelgrid_to_ds` function which already deals with a vertex dataset.

rubencalje:
Yes , `modelgrid_to_vertex_d` can probably be removed. It could be used to add modelgrid-information to an existing Dataset (and not create a new Dataset like `modelgrid_to_ds`. I do not think we use it anywhere, and  I do not see a use case for it.

PR #378, opened 2024-10-21
Small improvements

dbrakenhoff:
- Load ds from binary model grid (e.g. model.disv.grb file)
- Improve pyproj CRS initialization
- Add get_area() method using shoelace formula
- allow plotting modelgrid with or without rotation.


rubencalje:
Looks good. Would `add_area_to_ds` not be a better method name, instead of `get_area`?

OnnoEbbens:
I think it is better to return the area and not add the area to the dataset. Like we do in `get_idomain`. Also discussed in this issue https://github.com/gwmod/nlmod/issues/254

rubencalje (approved):

PR #380, opened 2024-10-23
Add BGT bronhouder names 

martinvonk:


bdestombe:
What about adding a test that checks whether the dictionary matches the current one found on het kadaster?

The following code compares the top row of the table presented on https://www.kadaster.nl/-/bgt-bronhoudercodes with the current implemented version

```python
import requests
from lxml import html

# The URL of the page
url = 'https://www.kadaster.nl/-/bgt-bronhoudercodes'

# Send a GET request to fetch the raw HTML content
response = requests.get(url)

# Parse the HTML content
tree = html.fromstring(response.content)

# Use the XPath to find the desired text
xpath = '//*[@id="collapse1"]/div/ul/li[1]/text()'
text = tree.xpath(xpath)[0]

assert text == "Versie 2024: 1-1-2024", "Bronhoudercodes are not up to date. Update `get_bronhouder_names()`."
```

martinvonk:
> What about adding a test that checks whether the dictionary matches the current one found on het kadaster?

Good point. I can add a check for that. Is there a formal protocol to do this in nlmod? For the knmi data platform the issue is the same I think with an open api key that can expire.

Another option would be to download and read the excel / ods file, but I'd advise against that since it looks unstructured.

martinvonk:
In my previous comment, I was a bit confused if you wanted a check in the get_bronhouder_names or a test. But I now added a test to assert if the date from the last url/file on the kadaster website is from 2024.

bdestombe:
Perfect! Thank you for the consideration. 

The reason to opt for only checking the year opposed to whether it is the most current version is because, an update once a year is sufficient?

martinvonk:
> The reason to opt for only checking the year opposed to whether it is the most current version is because, an update once a year is sufficient?

Yes, generally there is no big update. But you're right; in 2022 the file was updated twice. So I can also check the date as well.

rubencalje (approved):

PR #381, opened 2024-10-24
use local file if ellips server returns a HTTPError

OnnoEbbens:
Added a geojson with the ahn tiles and a function `_get_tiles_from_file` to prevent HTTPErrors when retrieving the tiles from the ellips server.

I considered removing the `_get_tiles_ellipsis` function and always use the geojson but then updates and changes to this file will go unnoticed, that's why I left the original function and added a try/except.


bdestombe:
> Added a geojson with the ahn tiles and a function `_get_tiles_from_file` to prevent HTTPErrors when retrieving the tiles from the ellips server.
> 
> I considered removing the `_get_tiles_ellipsis` function and always use the geojson but then updates and changes to this file will go unnoticed, that's why I left the original function and added a try/except.

What about always using the local geojson and adding a test that checks whether the local geojson is the same as the one served from the ellips server?

rubencalje:
Ah, I merged this PR, but I may have been a bit too quick, sorry.

The columns in EllipsisDrive_index_fancy.geojson are different from the columns you normally retrieve using `_get_tiles_ellipsis`: for example "AHN3 maaiveldmodel (DTM) ½m" instead of "AHN3_05M_M".

I will add a fix and implement the suggestion by @bdestombe.

rubencalje:
> Ah, I merged this PR, but I may have been a bit too quick, sorry.
> 
> The columns in EllipsisDrive_index_fancy.geojson are different from the columns you normally retrieve using `_get_tiles_ellipsis`: for example "AHN3 maaiveldmodel (DTM) ½m" instead of "AHN3_05M_M".
> 
> I will add a fix and implement the suggestion by @bdestombe.

I fixed the AHN in PR #383. By default, we now use the geojson-file for the metadata of the tiles, and I added a test that checks if the geojson-file is upto-date.

PR #385, opened 2024-11-13
Release v0.9.0

dbrakenhoff:


rubencalje:
The NHI-GWO database does not return data for Brabant Water anymore, which is the reason this tests in the PRs was failing. I sent a mail to the NHI helpdesk, with a question why this is the case. For now, maybe we can change the test to a well field of another drinking water company, for example Vitens.

nlmod/read/bgt.py
```diff
@@ -321,3 +326,517 @@ def get_bgt_layers():
...
+    -----
+        The bronhouder names are retrieved from
+        https://www.kadaster.nl/-/bgt-bronhoudercodes. The `Toelichting` sheet
+        in the .ods file gives the changes compared to the old file. If changes
+        are made, please add them manually to the bgt_bronhouder_names
+        dictionary. A test is added for to check if the dictionary is up to
+        date with the latest file from the Kadaster up to date with the .ods
+        file from 2024-01-01.
+    """
+    bgt_bronhouder_names = {
```
OnnoEbbens, line 349: I would prefer to put 300 lines of data in a separate data file not in a function

nlmod/version.py
```diff
@@ -1,7 +1,7 @@
 from importlib import metadata
 from platform import python_version
 
-__version__ = "0.8.1"
+__version__ = "0.8.2.dev0"
```
OnnoEbbens, line 4:
```suggestion
__version__ = "0.9.0"
```

OnnoEbbens (changes requested):
Some comments from me, I can also fix them if you want.

nlmod/data/geotop/REF_GTP_LITHO_CLASS.xlsx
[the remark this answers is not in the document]
OnnoEbbens: I see this PR message now (https://github.com/gwmod/nlmod/pull/361) and the files are only 26 kb so not something to worry about.

nlmod/read/bgt.py
OnnoEbbens: fixed

OnnoEbbens (approved):

Issue #310, opened 2023-12-27
calculate_resistance calculates resistance between layers


rubencalje:
Right now `calculate_resistance` calculates the resistance between layers (also see issue #134), and not the resistance of layers themselves (if that is correct english...).

I propose to rename the current `calculate_resistance` to `calculate_resistance_between_layers`, and create a new `calculate_reistance` with the formula:

`c = nlmod.layers.calculate_thickness(ds) / ds['kh']`.

Any objections?

rubencalje:
Or add an optional argument to `calculate_resistance` named `between_layers`, which is set to `False` by default (which would change the current outcome of `calculate_resistance`). I prefer this option.

PR #386, opened 2024-11-13
Add between_layers to calculate_resistance, and set to False

rubencalje:
Solves issue #310. This PR changes the outcome of `calculate_resistance`, when `between_layers` is set to False. This will be the new default in the future as well. When `between_layers` is not specified (which is the default), the user will see the following warning:

`The default of between_layers=True in calculate_resistance is deprecated and will be changed to False in a future version of nlmod. Pass between_layers=True to retain current behavior or between_layers=False to adopt the future default and silence this warning.`

### DIFF
--- a/nlmod/dims/layers.py
+++ b/nlmod/dims/layers.py
@@ -121,10 +121,11 @@ def calculate_transmissivity(
     return T
 
 
-def calculate_resistance(ds, kv="kv", thickness="thickness", top="top", botm="botm"):
-    """Calculate vertical resistance (c) between model layers from the vertical
-    conductivity (kv) and the thickness. The resistance between two layers is assigned
-    to the top layer. The bottom model layer gets a resistance of infinity.
+def calculate_resistance(
+    ds, kv="kv", thickness="thickness", top="top", botm="botm", between_layers=False
+):
+    """Calculate vertical resistance (c) of model layers from the vertical conductivity
+    (kv) and the thickness.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


nlmod/dims/layers.py
```diff
@@ -142,6 +143,12 @@ def calculate_resistance(ds, kv="kv", thickness="thickness", top="top", botm="bo
     botm : str, optional
         name of data variable containing bottoms, only used to calculate thickness if not
         available in dataset. By default "botm"
+    between_layers : bool, optional
+        If True, calculate the resistance between the layers, which MODFLOW uses to
+        calculate the flow. The resistance between two layers is then assigned to the
+        top layer, and the bottom model layer gets a resistance of infinity.
+        If False, calculate the resistance of the layers themselves. The defauls is
```
dbrakenhoff, line 150: typo

nlmod/dims/layers.py
```diff
@@ -121,10 +121,11 @@ def calculate_transmissivity(
     return T
 
 
-def calculate_resistance(ds, kv="kv", thickness="thickness", top="top", botm="botm"):
-    """Calculate vertical resistance (c) between model layers from the vertical
-    conductivity (kv) and the thickness. The resistance between two layers is assigned
-    to the top layer. The bottom model layer gets a resistance of infinity.
+def calculate_resistance(
+    ds, kv="kv", thickness="thickness", top="top", botm="botm", between_layers=False
```
dbrakenhoff, line 125: maybe set to None, warn users that to get the default they need to set between_layers=False, and the future default will be True?

dbrakenhoff (commented):
See comment

rubencalje:
Test fail unrelated to this PR

PR #388, opened 2024-11-13
add RWS bathymetry to nlmod.read.rws

dbrakenhoff:
- add method to get configuration
- add method to download index (geodataframe)
- add method to download bathymetry data as xr.DataArray
- add tests

### DIFF
--- a/nlmod/dims/grid.py
+++ b/nlmod/dims/grid.py
@@ -132,7 +132,7 @@ def get_icell2d_from_xy(x, y, ds, gi=None, rotated=True):
 
     Raises
     ------
-
+    ValueError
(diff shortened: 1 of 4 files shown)


rubencalje (approved):
Great addition! I would change the following minor points:
- I would change the default resolution to 20m (unless there is a good reason for the fine resolution)
- I would add an optional parameter `config` to `get_bathymetry_gdf` and `get_bathymetry`, with a value of None that is replaced by the output of `get_gdr_configuration` when None. IN this way, when the url changes, the user can edit the configuration without a new nlmod version.

dbrakenhoff:
test fails unrelated.

Issue #389, opened 2024-11-29
Cache is never used because hashes do not match


dbrakenhoff:
At the moment I cannot use the caching functionality in one of my projects. Confirmed to happen for REGIS and AHN datasets. 

For some reason the hashes never match, causing the cache to be invalidated. However, when I try reproducing this in a separate minimal example, the caching is working fine. 

If anyone has any ideas, I'd love to hear them, otherwise as my investigation continues I will post updates here.

This is working fine...
```python
import nlmod

nlmod.util.get_color_logger("DEBUG")

cachedir = "."
extent = [100_000, 101_000, 400_000, 401_000]

regis_ds = nlmod.read.regis.get_regis(extent, cachedir=cachedir, cachename="regis")
regis_ds = nlmod.read.regis.get_regis(extent, cachedir=cachedir, cachename="regis")
```

Versions:
```
Python version     : 3.11.10
NumPy version      : 1.26.4
Xarray version     : 2024.9.0
Matplotlib version : 3.9.2
Flopy version      : 3.9.0.dev1

nlmod version      : 0.9.0
```

OnnoEbbens:
I tried the following in a GH codespace (Debian GNU/Linux 12 (bookworm):
1. run script 1 (see below)
2. restart the kernel
3. run script 2 (see below)
4. compare the hashes created in both scripts (see below)

What I get from this:
- Reading the same cached netcdf file after restarting the kernel results in a different hash. I visually compared the two datasets (dimensions, coordinates, data variables and attributes) and I don't see any differences.
- Same for reading the same pickled dataset.

I don't know why the hashes are different but I guess it is not something we can easily solve.

The intention of the hash is to check if the pickled function arguments where created together with the cached netcdf file. In other words if the .pklz and the .nc file belong together. This is a nice to have check but the cache will work without. So I would propose to disable the hash check from the cache until we found a solution for this.

packageversions used:
python 3.11.10
nlmod 0.9.0
xarray 2023.6.0
dask 2024.12.0

script 1
```
import nlmod
import pickle
import xarray as xr
import dask

# instellingen voor de logging
nlmod.util.get_color_logger("INFO")

cachedir ='.'
extent = [204800, 205000, 438000, 438200]

#%%
# get regis dataset
regis_ds = nlmod.read.regis.get_regis(extent, cachedir=cachedir, cachename="regis")
hash_orig = dask.base.tokenize(regis_ds)

# write to pickle
[...]

PR #393, opened 2024-12-16
Temporary disable hash

OnnoEbbens:
As discussed in #389 the hash check is temporary disabled


dbrakenhoff (commented):
It seems a fix has been submitted? And perhaps it would be better if can include a hash=True kwarg to the `cache_netcdf()` function?

dbrakenhoff:
Maybe close this PR, and optionally modify #395 to include the option to turn of hash checking?

OnnoEbbens:
Yes, that's is a good idea. for now I close this and will make an issue of optionally hash checking

PR #395, opened 2024-12-18
[cache] Validate cached file with hashlib instead of dask tokenize

bdestombe:


dbrakenhoff:
Does this make #393 unnecessary?

bdestombe:
Yes.

It works in my script and I tested it with the code of Onno. 
`hashlib_hash2 = hashlib.sha256(open(fp_cache, 'rb').read()).hexdigest()` the hash returned by the first script is the same as the one returned by the second script

dbrakenhoff (approved):

Issue #390, opened 2024-12-03
Replace MODPATH by PRT


rubencalje:
RIght now there is support for MODPATH in NLmod. However, we use a provisional version downloaded from
https://github.com/MODFLOW-USGS/modpath-v7/tree/develop/msvs/bin_PROVISIONAL

In our setup of generating the documentation this version now gives an error (in linux), for which we have not found a solution:

> OSError: [Errno 8] Exec format error: '/home/docs/checkouts/readthedocs.org/user_builds/nlmod/envs/stable/lib/python3.10/site-packages/nlmod/bin/mp7_2_002_provisional'

I think we should replace support for MODPATH with support for the Particle Tracking (PRT) Model for MODFLOW 6, described on the MODFLOW 6 page of the USGS:

> The Particle Tracking (PRT) Model for MODFLOW 6 calculates three-dimensional, advective particle trajectories in flowing groundwater. The PRT Model is designed to work with the Groundwater Flow (GWF) Model (Langevin and others, 2017) for MODFLOW 6, which simulates transient, three-dimensional groundwater flow. The PRT Model replicates much of the functionality of MODPATH 7 and offers support for a much broader class of unstructured grids. The PRT model uses the same spatial discretization used by the GWF Model, which may be represented using either a structured (DIS) or an unstructured (DISV) grid. The PRT Model can be run in the same simulation as the associated GWF Model or in a separate simulation that reads previously calculated flows from a binary budget file. The current version of the PRT Model does not support grids of DISU type, tracking of particles through advanced stress package features such as lakes or streams reaches, or exchange of particles between PRT models. The PRT Model simulates the forward tracking of particles; however, backward particle tracking is possible using a procedure described here for reversing simulated flows from a previous GWF simulation and using those flows as input to PRT. 

I think it is good to aim for support in 2025.

PR #398, opened 2024-12-24
Add PRT support

dbrakenhoff:
First attempt at addressing #390. Still somewhat under development on the flopy side, so creating a draft PR for now. 

Adds
- PRT submodule with all relevant packages (except FMI for now)
- EMS solver
- improve some of the modpath utilities to be applicable for PRT models (i.e. don't need a modpath model).

Todo:
- [ ] add nlmod FMI package?
- [ ] utilities to write PRP data (without having to use modpath tools which currently need to perform lots of coordinate transformations)
- [ ] Make notebook slightly more interesting

### DIFF
--- a/docs/examples/18_particle_tracking_prt.ipynb
+++ b/docs/examples/18_particle_tracking_prt.ipynb
@@ -1,0 +1,168 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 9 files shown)


dbrakenhoff:
Requires latest develop version of flopy (or maybe even one of the open PRT PRs) to work.

martinvonk (approved):
I have an example for disv, but I'll add that later. But this works so very happy :)

PR #399, opened 2024-12-24
Add cross-section plot utils

dbrakenhoff:
- add inset_map
- add xsec_line_and_labels


rubencalje:
Can you add a small test so we can see the result of the new methods?

dbrakenhoff:
So demanding... ;). 

I added a test in `test_011_dcs.py`:

image

And an example in `13_plot_methods.ipynb`:

image

PR #402, opened 2025-01-15
Parsing of 'c' value as float instead of timedelta[ns]

MattBrst:
By default the 'c' value was read as a timedelta[ns]. So a value of -9999 days was shown as -863913600000000000 nanoseconds and therefore not caught by the NaN replacement in the following lines.

### DIFF
--- a/nlmod/read/regis.py
+++ b/nlmod/read/regis.py
@@ -112,10 +112,10 @@ def get_regis(
         included in the model. the Default is "AKc" which is the bottom
         layer of regis. call nlmod.read.regis.get_layer_names() to get a list
         of regis names.
-    variables : tuple, optional
-        a tuple of the variables to keep from the regis Dataset. Possible
-        entries in the list are 'top', 'botm', 'kD', 'c', 'kh', 'kv', 'sdh' and
-        'sdv'. The default is ("top", "botm", "kh", "kv").
+    variables : tuple or list, optional
+        The variables to keep from the regis Dataset. Possible entries in the list are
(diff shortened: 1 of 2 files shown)


dbrakenhoff:
Nice catch! Maybe even better to confirm the dtype of the "c" variable before doing the conversion? This feels like one of those things that could change without us knowing. 

So maybe we could add: `if "c" in variables and ds["c"].dtype.type is np.timedelta64:`

dbrakenhoff (approved):
Nice, thanks for the PR! Will merge after tests complete.

rubencalje:
Thanks for the PR!

It might be better to add `decode_times=False` to `xr.open_dataset()`. In this way, no data is read as datetimes. I do not believe there is any time-data in the regis-dataset. We may also remove `decode_coords="all"` from `xr.open_dataset()`, as I do not remember why we added this in the first place:

`ds = xr.open_dataset(REGIS_URL, decode_times=False) `

rubencalje:
> Thanks for the PR!
> 
> It might be better to add `decode_times=False` to `xr.open_dataset()`. In this way, no data is read as datetimes. I do not believe there is any time-data in the regis-dataset. We may also remove `decode_coords="all"` from `xr.open_dataset()`, as I do not remember why we added this in the first place:
> 
> `ds = xr.open_dataset(REGIS_URL, decode_times=False) `

Hmm, I changed this 4 months ago: 555e0f060cca170026e68f614abfb628bd9d40e0. Not sure why anymore...

dbrakenhoff (approved):
Nice! Looks a lot more robust now :).

PR #407, opened 2025-02-20
Bofek improvements

OnnoEbbens:
I got an error using the previous `get_gdf_bofek` function because it automatically writes to the NLMOD_DATADIR. My script did not have write permissions to this directory. Therefore, I added the option to specify a directory yourself.

I also tried very hard to do everything in memory so you don't have to write data to the disk. This involved many steps:
1. read data from URL
2. extract zipped data, this return another 7zip folder
3. extract only the geodatabase files from this folder
4. create a geodataframe from the geodatabase files

I did succeed to do steps 1 and 2 in-memory but I did not succeed to do steps 3 and 4 in-memory. Adding steps 1 and 2 unfortunately removed the progress bar and I am not sure how to get it back. @dbrakenhoff: do you know how to do this?

For future reference: I almost managed to complete steps 3 and 4 using a private method from py7zr. I leave the code here just in case anybody feels like trying this:

```
with py7zr.SevenZipFile(BytesIO(unzipped), mode="r") as a:
    out= a._extract(path=None, return_dict=True, targets=["GIS/BOFEK2020_bestanden/BOFEK2020.gdb"], recursive=True)
    # out= a.extract(path=None, targets=["GIS/BOFEK2020_bestanden/BOFEK2020.gdb"], recursive=True)

zip_buffer = BytesIO()

with ZipFile(zip_buffer, "a", ZIP_DEFLATED, False) as zip_file:
    for file_name, data in out.items():
        zip_file.writestr(file_name, data.getvalue())


with ZipMemoryFile(zip_buffer.getvalue()) as memfile:
    with memfile.open() as src:
        resultaat = gpd.GeoDataframe(src)
```

### DIFF
--- a/nlmod/read/bofek.py
+++ b/nlmod/read/bofek.py
@@ -1,19 +1,19 @@
 import logging
 import shutil
 import zipfile
+from io import BytesIO
 from pathlib import Path
 
 import geopandas as gpd
 import requests
-from tqdm.auto import tqdm
 
-from nlmod import NLMOD_DATADIR, cache, dims, util
+from nlmod import NLMOD_DATADIR, cache, util
 
 logger = logging.getLogger(__name__)
 
 
 @cache.cache_pickle
-def get_gdf_bofek(ds=None, extent=None, timeout=3600):
+def get_gdf_bofek(extent, dirname=None, timeout=3600):
     """Get geodataframe of bofek 2020 wihtin the extent of the model.
 
     It does so by downloading a zip file (> 100 MB) and extracting the relevant
@@ -22,10 +22,11 @@ def get_gdf_bofek(ds=None, extent=None, timeout=3600):
 
     Parameters
     ----------
-    ds : xr.DataSet, optional
-        dataset containing relevant model information. The default is None.
-    extent : list, tuple or np.array, optional
-        extent xmin, xmax, ymin, ymax. Only used if ds is None. The default is None.
+    extent : list, tuple or np.array
+        extent xmin, xmax, ymin, ymax.
+    dirname : str, optional
+        directory name for the bofek2020 files. If None the NLMOD_DATADIR is used to
+        store the data. The default is None.
     timeout : int, optional
         timeout time of request in seconds. Default is 3600.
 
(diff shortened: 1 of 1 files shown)


dbrakenhoff:
Do we really want to support loading all this in memory, especially since it seems quite challenging to do so fully? In which case, since we still need to write data somewhere, we might as well write the rest as well.

Good idea to pass the output directory. And instead of using NLMOD_DATADIR by default maybe this should even be a required argument. And then I would just keep the original code (with the progressbar).

dbrakenhoff:
Having said that, you should be able to write in the same way to a BytesIO object:

```python
# show download progress
total_size = int(r.headers.get("content-length", 0))
block_size = 1024
file = BytesIO()
with tqdm(
    total=total_size, unit="B", unit_scale=True, desc="Downloading BOFEK"
) as progress_bar:
    for data in r.iter_content(block_size):
        progress_bar.update(len(data))
        file.write(data)
```

OnnoEbbens:
I agree and removed the in-memory part. Also restructured a bit. 

I also tried your other suggestion to write the file in memory combined with tqdm but it did not work for me. I get an error (`Bad7zFile: not a 7z file`) when I try to read the file using `py7zr.SevenZipFile` using this code:

```
# show download progress
    total_size = int(r.headers.get("content-length", 0))
    block_size = 1024
    file_unzipped = BytesIO()
    with tqdm(
        total=total_size, unit="B", unit_scale=True, desc="Downloading BOFEK"
    ) as progress_bar:
        for data in r.iter_content(block_size):
            progress_bar.update(len(data))
            file_unzipped.write(data)

    # extract geodatabase from 7z
    with py7zr.SevenZipFile(file_unzipped, mode="r") as z:
        z.extract(
            targets=["GIS/BOFEK2020_bestanden/BOFEK2020.gdb"],
            path=dirname,
            recursive=True,
        )
```

dbrakenhoff:
> I get an error (Bad7zFile: not a 7z file) when I try to read the file using py7zr.SevenZipFile using this code:

I opened the result with `Zipfile` and not 7z which seemed to work...?

OnnoEbbens:
Ah yes that works. I updated my code to re-add the in-memory part with the progress bar. I like it better like this because it is less code and it run in 28 seconds while the previous one took 34 seconds.

dbrakenhoff (approved):
Nice!

Issue #406, opened 2025-02-13
combine_layers_ds and tuples with length < 2


Arjanvw:
When having tuples of length one in "combine_layers", combine_layers_ds throws an error. 

Maybe specify in function description or add line of code to handle:
combine_layers = [t for t in combine_layers if len(t) > 1]

rubencalje:
Fixed since version 0.11.0

PR #410, opened 2025-02-24
Improve combine layers + some tiny fixes

dbrakenhoff:
Improve combine_layers:
- accept layer names and layer indices  for combine_layers (as both dict and list)
- raise a useful error message when layers are non-consecutive
- better docstrings
- remove unnecessary ordereddict
- deal with single entries in input (#406)

Other fixes:
- tiny change in xsec label placement in add_xsec_line_and_labels
- add xmargin/ymargin to get_extent to expand/shrink extent by some amount in x and y

### DIFF
--- a/nlmod/dims/layers.py
+++ b/nlmod/dims/layers.py
@@ -616,28 +616,52 @@ def combine_layers_ds(
     # calculate new tops/bots
     logger.info("Calculating new layer tops and bottoms...")
 
-    if "layer" in ds["top"].dims:
+    if "layer" in ds[top].dims:
         msg = "Top in ds has a layer dimension. combine_layers_ds will remove the layer dimension from top in ds."
         logger.warning(msg)
     else:
         ds = ds.copy()
-        ds["top"] = ds["botm"] + calculate_thickness(ds)
+        ds[top] = ds[bot] + calculate_thickness(ds)
 
     if isinstance(combine_layers, dict):
+        # remove single layer entries if they exist:
+        combine_layers = {k: v for k, v in combine_layers.items() if len(v) > 1}
         new_layer_names = combine_layers.keys()
-        combine_layers = [
-            tuple(np.where(ds.layer.isin(x))[0]) for x in combine_layers.values()
+        combine_layers_integer = [
+            tuple(np.where(ds.layer.isin(x))[0]) if isinstance(x[0], str) else x
+            for x in combine_layers.values()
         ]
-        # make sure there are no layers in between
-        assert np.all([(np.diff(x) == 1).all() for x in combine_layers])
     else:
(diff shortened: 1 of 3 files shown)


nlmod/dims/layers.py
```diff
@@ -613,28 +641,52 @@ def combine_layers_ds(
     # calculate new tops/bots
     logger.info("Calculating new layer tops and bottoms...")
 
-    if "layer" in ds["top"].dims:
+    if "layer" in ds[top].dims:
         msg = "Top in ds has a layer dimension. combine_layers_ds will remove the layer dimension from top in ds."
```
OnnoEbbens, line 645:
```suggestion
        msg = f"datavar {top} has a layer dimension. combine_layers_ds will remove the layer dimension from {top} in ds."
```

nlmod/dims/layers.py
```diff
@@ -678,7 +730,7 @@ def combine_layers_ds(
     ds_combine = xr.Dataset(da_dict, attrs=attrs)
 
     # remove layer dimension from top again
-    ds = remove_layer_dim_from_top(ds, inconsistency_threshold=0.001)
+    ds = remove_layer_dim_from_top(ds, inconsistency_threshold=1e-3)
```
OnnoEbbens, line 733: this is not the one that is returned, so I guess the layer dimension is not removed

nlmod/dims/layers.py
```diff
@@ -401,14 +400,17 @@ def layer_combine_top_bot(ds, combine_layers, layer="layer", top="top", bot="bot
         if i in np.concatenate(combine_layers):
             # get indices of layers
             c = combine_layers[icomb]
+            old_names = ds.layer.isel(layer=list(c)).layer.to_numpy().tolist()
```
OnnoEbbens, line 403:
```suggestion
            old_names = ds.layer.isel(layer=list(c)).to_numpy().tolist()
```

nlmod/dims/layers.py
OnnoEbbens: At the moment the opposite is True, see my next comment.

OnnoEbbens (changes requested):
Nice fixes! I think there is some inconsistency with removing the layer dimension from the top. It was already there before this PR. I think it is easily fixed, although it might break some old scripts.

nlmod/dims/layers.py
dbrakenhoff: In place maybe?

dbrakenhoff: I think the code was modified, but the message never was, so I've changed it now. The layer dim is actually removed from the top in the returned dataset.

OnnoEbbens (approved):

PR #413, opened 2025-02-28
Use hashlib instead of dask.base.tokenize

dbrakenhoff:
Not sure if this PR should be merged, but I wanted to publish this code somewhere. 

This moves all hashing checks away from `dask.base.tokenize` and uses hashlib instead. Two methods are introduced, `hash_xarray_coords()` for checking the coordinates and `hash_xarray_data_vars()` for the data variables. This also gives more fine-grained control over which data to compute the hash for through the `include_metadata` kwarg. The default is False, because I wasn't getting corresponding hashes for my REGIS test dataset when including the metadata.  

With these changes my cached copy of  AHN survives a kernel restart without having to override caching settings introduced in #412. Was it worth it? I'm not sure :P. 

In my opinion this adds even more complexity to the whole caching thing (which was already complicated) so I would actually prefer just to do an explicit coordinate check and ignore all the hash comparisons (great idea on paper, but hard to configure in practice). 

So, curious to hear your thoughts on this.

### DIFF
--- a/nlmod/cache.py
+++ b/nlmod/cache.py
@@ -1025,7 +1025,7 @@ def hash_xarray_data_vars(
         if include_metadata:
             # hash each coordinate separately
             coord_hashes = {}
-            for coord_name, coord in da.coords.items():
+            for coord_name, coord in sorted(da.coords.items()):
                 coord_hashes[coord_name] = hash_xarray_coords(
                     coord, include_metadata=False
                 )
--- a/nlmod/cache.py
+++ b/nlmod/cache.py
@@ -940,12 +940,12 @@ def _explicit_dataset_coordinate_comparison(ds_in, ds_cache):
 class NumpyEncoder(json.JSONEncoder):
     """Special json encoder for numpy types."""
 
-    def default(self, obj):
-        if isinstance(obj, np.integer):
-            return int(obj)
-        elif isinstance(obj, np.floating):
-            return float(obj)
-        return json.JSONEncoder.default(self, obj)
+    def default(self, o):
+        if isinstance(o, np.integer):
+            return int(o)
+        elif isinstance(o, np.floating):
+            return float(o)
+        return json.JSONEncoder.default(self, o)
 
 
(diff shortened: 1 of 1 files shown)


bdestombe:
> In my opinion this adds even more complexity to the whole caching thing (which was already complicated) so I would actually prefer just to do an explicit coordinate check and ignore all the hash comparisons (great idea on paper, but hard to configure in practice).

What do you mean by doing an explicit coordinate check? Excluding the metadata? I'm all up for excluding the metadata of the coords and datavars as none of it is used in any of the nlmod functions. I like your default. Specific attributes of the ds are relevant but these are passed to ds_contains directly.

Thanks for all the hard work!

nlmod/cache.py
```diff
@@ -915,3 +935,115 @@ def _explicit_dataset_coordinate_comparison(ds_in, ds_cache):
...
+
+    Returns
+    -------
+    str
+        The hexadecimal hash string
+    """
+    # get the raw bytes from the numpy array values
+    combined_bytes = b""
+    if isinstance(ds, xr.Dataset):
+        data_arrays = [ds[da] for da in ds.data_vars]
```
bdestombe, line 1016: Do we have to worry about the ds being sorted?

bdestombe (changes requested):
I'm a bit worried about the sorting order of the datavars. Sorting the list of da's by name alphabetically would solve it

nlmod/cache.py
dbrakenhoff: Good question, it would be prudent to sort to make sure, as you said.

bdestombe:
And maybe also sort the coordinates and the attributes?

dbrakenhoff:
> And maybe also sort the coordinates and the attributes?

Done, attributes were already sorted with `sort_keys=True`

bdestombe (approved):

bdestombe:
@David, @OnnoEbbens Would you be okay merging this? Currently, cache is not working for me and with these changes it does: #389

PR #415, opened 2025-02-28
allow passing multiple lake observation types as a list of str

dbrakenhoff:
- e.g. `obs_type=["stage", "lak"]`


rubencalje (approved):
Approve. Would be nice if the docstring is updated, and the same changes was applied to obs_type_gwt.

PR #422, opened 2025-03-08
allow getting layer top/botm from ds

dbrakenhoff:
- accessing gwf.modelgrid arrays is insanely slow for some reason.
- add a different strt computation when strt=None. Default is now to use model surface level instead of 0.0.

### DIFF
--- a/nlmod/gwf/wells.py
+++ b/nlmod/gwf/wells.py
@@ -137,6 +137,7 @@ def maw_from_df(
     aux=None,
     boundnames=None,
     ds=None,
+    silent=False,
     **kwargs,
 ):
     """Add a Multi-Aquifer Well (MAW) package based on input from a (Geo)DataFrame.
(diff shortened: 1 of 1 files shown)


nlmod/gwf/wells.py
```diff
@@ -199,11 +200,26 @@ def maw_from_df(
...
+                    gwf.dis.top[irow["cellid"]]
+                    if ds is None
+                    else ds.top.values[irow["cellid"]]
+                )
+            else:
+                wstrt = (
+                    gwf.dis.top[irow["cellid"][0], irow["cellid"][1]]
+                    if ds is None
+                    else ds.top.values[irow["cellid"][0], irow["cellid"][1]]
+                )
```
rubencalje, line 219: Can you make regular if-statements from these lines? That would be easier to read than these one-line if statements that rae sprread over multuiple lines.

rubencalje: Can you change the docstring of strt, and inform the user with what value strt is replaced when strt is None?

nlmod/gwf/wells.py
```diff
@@ -199,11 +200,26 @@ def maw_from_df(
     perioddata = []
 
     iw = 0
-    for index, irow in df.iterrows():
+    for index, irow in tqdm(df.iterrows(), total=len(df), desc="Adding MAW wells"):
```
rubencalje, line 203: maybe add a optional 'silent' argument to maw_from_df (with a default value of False), which is passed to tqdm as a 'disable` argument. I think we have similar uses in other parts of nlmod.

rubencalje (approved):
Great! One code-formatting comment, one docstring-comment, and one extra argument suggestion. But can also be merged without any changes if you want to get rid of it.

PR #424, opened 2025-03-12
Add rename_layers_to_old_names to get_regis

rubencalje:
See issue #418. This PR handles option 3 mentioned in the comments: it renames the regis layers to their original names as soon as we download REGIS.  On march 18 we can test if this PR performs as expected.


dbrakenhoff (approved):
Looks good, let's see if it works on March 18th :)

martinvonk:
> Looks good, let's see if it works on March 18th :)

spannund

PR #430, opened 2025-03-18
Release v0.9.3

rubencalje:
This release is made because of a change in the layer names of REGIS since today (march 18 2025). The layer names are renamed to their old names for now, so existing model-scripts behave the same as before. In a next mayor version, we may choose to adopt the new layer names.


rubencalje:
Codacy issues can be ignored.

Issue #429, opened 2025-03-17
Merging AHN tiles takes forever


martinvonk:
When downloading the AHN5 (and 4) for a relatively large extent and at 05m scale.

```python
extent = [101_000, 120_000, 412_000, 432_000]
ahn = nlmod.read.ahn.get_ahn5(extent=extent, identifier="AHN5_05M_M")
```
results in
```
Downloading tiles of AHN5_05M_M: 100%|██████████| 20/20 [00:03<00:00,  6.66it/s]
INFO:rasterio.merge.merge:Skipping source: src=<rioxarray.merge.RasterioDatasetDuck object at 0x7f05ad0dc2d0>, bounds=(105000.0, 425000.0, 110000.0, 431250.0)
INFO:rasterio.merge.merge:Skipping source: src=<rioxarray.merge.RasterioDatasetDuck object at 0x7f05ad0dd4d0>, bounds=(115000.0, 425000.0, 120000.0, 431250.0)
INFO:rasterio.merge.merge:Skipping source: src=<rioxarray.merge.RasterioDatasetDuck object at 0x7f05ad07c150>, bounds=(110000.0, 431250.0, 115000.0, 437500.0)
INFO:rasterio.merge.merge:Skipping source: src=<rioxarray.merge.RasterioDatasetDuck object at 0x7f05ad07fdd0>, bounds=(115000.0, 431250.0, 120000.0, 437500.0)
INFO:rasterio.merge.merge:Skipping source: src=<rioxarray.merge.RasterioDatasetDuck object at 0x7f05ad07fa90>, bounds=(100000.0, 418750.0, 105000.0, 425000.0)
INFO:rasterio.merge.merge:Skipping source: src=<rioxarray.merge.RasterioDatasetDuck object at 0x7f05ad0dc610>, bounds=(115000.0, 418750.0, 120000.0, 425000.0)
INFO:rasterio.merge.merge:Skipping source: src=<rioxarray.merge.RasterioDatasetDuck object at 0x7f05ad0e2190>, bounds=(110000.0, 425000.0, 115000.0, 431250.0)
INFO:rasterio.merge.merge:Skipping source: src=<rioxarray.merge.RasterioDatasetDuck object at 0x7f05ad0a5750>, bounds=(110000.0, 406250.0, 115000.0, 412500.0)
[...]

PR #431, opened 2025-03-18
Add option to merge ahn tiles

rubencalje:


rubencalje:
So a lot of changes, @martinvonk . Can you explain what the purpose of these changes is? And I think we should use type hints everywhere in NLmod, or no type hints at all (which I like more).

martinvonk:
I tried to add some code to use the old `_download_and_combine_tiles` function to merge the separate tiles. This function used a memoryfile thingy, and the `merge` function instead of `merge_arrays` which might be a solution to #429. 

And yes, I added some type hints 👼. Originally because I was confused about the expected input for some functions. Later more, because I thought it is nice to be explicit in the function call which specific version (strings) need to be provided for an AHN request. Btw, there are already some [type hints in NLmod](https://github.com/gwmod/nlmod/blob/dev/nlmod/read/knmi_data_platform.py). I'm always in favor of more documentation instead of less.

martinvonk:
> So does it work? Can you add a test?

It does not work yet :) But I'll have a look to see if I can fix it.

Otherwise, I'll look if I can catch the logging info shown in #429, and throw an error message to use the return_tiles argument.

martinvonk:
Alright, I gave up on trying to catch the logger.info warning and raise a warning to not merge the ahn tiles.

Issue #432, opened 2025-03-19
read.regis.get_layer_names() still returns old names


bdestombe:
https://github.com/gwmod/nlmod/blob/82ffc0ec4099444272a3780a0ed645f43f98e2ec/nlmod/read/regis.py#L356

PR #433, opened 2025-03-24
Update regis.get_layer_names() to return pre 2.2.3 names

bdestombe:
#432

### DIFF
--- a/nlmod/read/regis.py
+++ b/nlmod/read/regis.py
@@ -396,7 +396,7 @@ def get_layer_names(ds=None, rename_layers_to_version_2_2_2=True):
     """
     if ds is None:
         ds = xr.open_dataset(REGIS_URL, decode_times=False, decode_coords=False)
-    
+
     regis_version = extract_version_from_title(ds.attrs["title"])
 
     layer_names = ds.layer.values.astype(str)
--- a/nlmod/read/regis.py
+++ b/nlmod/read/regis.py
@@ -1,11 +1,11 @@
 import datetime as dt
 import logging
 import os
-from packaging.version import parse as parse_version
 
 import numpy as np
 import pandas as pd
 import xarray as xr
+from packaging.version import parse as parse_version
 
 from .. import cache
 from ..dims.layers import calculate_thickness, remove_layer_dim_from_top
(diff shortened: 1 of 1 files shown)


bdestombe:
The failed test is unrelated to the changes proposed here. They are related to a firewall at nhi.nu. 
```zsh
url = 'https://gwo.nhi.nu/api/v1/well_filters/', page = 1, properties = []
params = {'format': 'csv', 'n_well_filters': 1000, 'page': 1, 'well__organization': 'Brabant Water'}
r = <Response [403]>
content = '<!DOCTYPE html>\r\n<html>\r\n<head>\r\n    <title>Web App - Unavailable</title>\r\n    <style type="text/css">\r\n   ...app you have attempted to reach has blocked your access.</p>\r\n        </div>\r\n    </div>\r\n</body>\r\n</html>\r\n'
```

nlmod/read/regis.py
```diff
@@ -163,18 +164,18 @@ def get_regis(
         msg = "No data found. Please supply valid extent in the Netherlands in RD-coordinates"
         raise (ValueError(msg))
 
-    # make sure layer names are regular strings
-    ds["layer"] = ds["layer"].astype(str)
+    regis_version = extract_version_from_title(ds.attrs["title"])
+
+    if rename_layers_to_version_2_2_2 and regis_version >= parse_version("2.2.3"):
+        pre223_layer_names = get_layer_names(rename_layers_to_version_2_2_2=True)
+        ds = ds.assign_coords({"layer": pre223_layer_names})
```
rubencalje, line 171: This downloads regis a second time, and can be tricky when ds is changed before this line (in future commits). Maybe we can pass `ds` to `get_layer_names` as an optional parameter. We can then just do:
`ds["layer"] = get_layer_names(ds=ds, rename_layers_to_version_2_2_2=rename_layers_to_version_2_2_2)`
and remove some lines of double code, which is then only in `get_layer_names`.

rubencalje (changes requested):
Great addition! I have one request to make it a bit more robust.

nlmod/read/regis.py
bdestombe: Thanks, great suggestion!

bdestombe:
Failing tests are not related again with the changes made here

rubencalje (approved):

Issue #437, opened 2025-03-27
Test `test_get_recharge_not_available` is taking ~800s


dbrakenhoff:
This test is causing our actions runs to take forever. 

The cause, it turns out, is that february 20th is missing from all KNMI rainfall stations... causing fill_missing_obs to run through all stations up to De Bilt to try and fill it, which it cannot do.

![Image](https://github.com/user-attachments/assets/111d19e0-4c6f-4206-81f6-287fae310417)

This will probably eventually be fixed by KNMI (perhaps we can notify them somehow?). But for now we should modify the test to be somewhat less sensitive to missing dates by picking a shorter period to run the test on.

PR #438, opened 2025-03-27
Fix slow test

dbrakenhoff:
Addresses #437 

Just realized, this change will only work if hydropandas is up to date :innocent: . Maybe try to pick period with partially available data instead.


dbrakenhoff:
This test will keep being super slow until we make a new release of hydropandas or force nlmod to use the dev version of hydropandas in its tests.

PR #447, opened 2025-04-17
add function to get modellayer from a well screen

OnnoEbbens:
This will make it easier to compare observations with model results


OnnoEbbens:
Failing test is due to rate limit exceeding

martinvonk:
I tried to do something similar lately. Not using both screen top and bot, only the average filter height. But I made the algorithm in such a way that it ignored the aquitards (`k`) and looked for the nearest aquifer.
```python
oseries: pd.DataFrame = ... # dataframe with oseries metadata
ds: xr.Dataset = ... # model dataset
head: xr.DataArray = nlmod.gwf.get_heads_da(ds)

oseries["z"] = (oseries["screen_bottom"] + oseries["screen_top"]) / 2
for name in oseries.index:
    h = nlmod.gwf.get_head_at_point(
        head, x=oseries.at[name, "x"], y=oseries.at[name, "y"], ds=ds
    )
    z = oseries.at[name, "z"]
    layers = ds["layer"].values
    z_layers = np.append(h["top"].values, h["botm"].values)
    z_layers = (z_layers[:-1] + z_layers[1:]) / 2

    def find_nearest_aquifer_name(z, layers_z, layer_names):
        id = np.argmin(np.abs(layers_z - z))
        layer = layer_names[id]
        while layer != "Hlc" and layer[-2] == "k":
            layers_z = np.delete(layers_z, id)
            layer_names = np.delete(layer_names, id)
            id = np.argmin(np.abs(layers_z - z))
            layer = layer_names[id]
        return layer

    layer = find_nearest_aquifer_name(z, z_layers, layers)
```

Maybe we can do that here as well? But smarter (using both screen top and bot).

dbrakenhoff:
Haven't yet checked your code @OnnoEbbens, but for the sake of it, here is another option I used recently, in this case a structured implementation though it should be simple enough to convert to work for both unstructured and structured. 

This returns a float index number that indicates where in a layer the elevation `z` is located, e.g. a screen_top resulting in 1.7 means your screen starts 70% of the way down in layer in layer 1. This allows you to post-process to get a single layer, e.g. average the two results and round down, or show all layers between screen_top and screen_bottom.

```python
import nlmod
import numpy as np
import pandas as pd
import xarray as xr


def get_layer_position_1d(zvec, z, left=-999, right=999):
    lay = np.interp(
        z,
        zvec[::-1],
        np.arange(len(zvec))[::-1],
        left=left,
        right=right,
    )
    return lay


def get_surfaces_da(ds, top="top", botm="botm", top_layer_name="MV"):
    return (
        xr.combine_nested(
            [ds[top].expand_dims({"layer": [top_layer_name]}), ds[botm]],
            concat_dim="layer",
        )
        .rename("surfaces")
        .rename({"layer": "surface"})
    )


def get_layer_position(df, z, surfaces, x="x", y="y"):
    laypos = pd.Series(index=df.index, dtype=float)
    for i, (idx, row) in enumerate(df.iterrows()):
        r, c = nlmod.grid.get_row_col_from_xy(row[x], row[y], surfaces)
        if len(surfaces.dims) >= 2:
            isurf = surf.isel(x=c, y=r)
        else:
            isurf = surfaces
        if isinstance(z, str):
            zval = row[z]
        else:
            zval = z[i]
        laypos.loc[idx] = get_layer_position_1d(isurf, zval)
    return laypos
```

Usage: 
```python
surf = get_surfaces_da(ds)
df["ilay_ftop"] = get_layer_position(df, "screen_top", surf)
df["ilay_fbot"] = get_layer_position(df, "screen_bottom", surf)
```

nlmod/dims/layers.py
```diff
@@ -2003,3 +2005,116 @@ def get_isosurface(da, z, value, input_core_dims=None, exclude_dims=None, **kwar
...
+    Parameters
+    ----------
+    ds : xarray.Dataset
+        xarray Dataset with with top and bottoms, can be structured or vertex.
+    screen_top : np.ndarray of shape(nobs)
+        collection of screen top values).
+    screen_bottom : np.ndarray of shape(nobs)
+        collection of screen bottom values. Should have the same length as screen_top
+        and xy.
+    xy : np.ndarray of shape(nobs, 3), optional
```
dbrakenhoff, line 2023: shape (nobs, 3)?

nlmod/dims/layers.py
```diff
@@ -2003,3 +2005,116 @@ def get_isosurface(da, z, value, input_core_dims=None, exclude_dims=None, **kwar
...
+    """
+    if (ds_obs['screen_top'] <= ds_obs['screen_bot']).any():
+        raise ValueError('screen top is equal to or below screen bottom')
+
+    # get model layers for screen top and bottom
+    ds_obs['modellayer_top'] = (dimname,), [np.argmax(ds_obs['screen_top'].values[i]>ds_obs['botm'].values[:,i]) for i in range(ds_obs.sizes[dimname])]
+    ds_obs['modellayer_top'] = xr.where(ds_obs['screen_top'] >= ds_obs['top'], np.inf, ds_obs['modellayer_top'])
+    ds_obs['modellayer_top'] = xr.where(ds_obs['screen_top'] <= ds_obs['botm'].isel(layer=-1), -np.inf, ds_obs['modellayer_top'])
+    ds_obs['modellayer_bot'] = (dimname,), [np.argmax(ds_obs['screen_bot'].values[i]>ds_obs['botm'].values[:,i]) for i in range(ds_obs.sizes[dimname])]
+    ds_obs['modellayer_bot'] = xr.where(ds_obs['screen_bot'] >= ds_obs['top'], np.inf, ds_obs['modellayer_bot'])
```
dbrakenhoff, line 2088: maybe negative inf for below the model?

nlmod/dims/layers.py
```diff
@@ -2003,3 +2005,116 @@ def get_isosurface(da, z, value, input_core_dims=None, exclude_dims=None, **kwar
...
+    ds_obs['modellayer_top'] = xr.where(mask, 0, ds_obs['modellayer_top'])
+    ds_obs['screen_top'] = xr.where(mask, ds_obs['top'], ds_obs['screen_top'])
+
+    # screen bot is below model botm but screen top is above model botm
+    mask = (ds_obs['modellayer_bot']==-np.inf) & ~(ds_obs['modellayer_top']==-np.inf)
+    ds_obs['modellayer_bot'] = xr.where(mask, ds_obs.sizes['layer']-1, ds_obs['modellayer_bot'])
+    ds_obs['screen_bot'] = xr.where(mask, ds_obs['botm'][-1], ds_obs['screen_bot'])
+
+    # combine modellayer_top and modellayer_bot to get modellayer
+    def get_thickest_mlayer(i):
```
dbrakenhoff, line 2102: write out `get_thickest_model_layer`? Or use something with `max_overlap`?

tests/test_009_layers.py
```diff
@@ -256,3 +257,18 @@ def test_remove_thin_layers():
...
+    ds = test_001_model.get_ds_from_cache("small_model")
+    xy = [[98900,489600],[98800,489500],[98980,489680],[98980,489680]]
+    screen_top = [10, -1, -35, 1000]
+    screen_bottom = [9, -20, -100, -1000]
+    modellayers = nlmod.layers.get_modellayers_screens(ds, screen_top, screen_bottom, xy=xy)
+    assert np.isnan(modellayers[0])
+    assert modellayers[1] == 1.0
+    assert modellayers[2] == ds.sizes['layer']-1
+
+    ds_ref = nlmod.grid.refine(ds, refinement_features=[])
```
dbrakenhoff, line 272: worth making this a fixture in another PR :innocent:

nlmod/dims/layers.py
```diff
@@ -2003,3 +2005,116 @@ def get_isosurface(da, z, value, input_core_dims=None, exclude_dims=None, **kwar
...
+
+def _get_modellayers_dsobs(ds_obs, dimname='no_obs'):
+    """Get modellayers from a dataset of observation point data
+
+    Parameters
+    ----------
+    ds_obs : xr.Dataset
+        typically a subset of a model dataset with only data for observations.
+    dimname : str, optional
+        name of the observation dimension. Usually 'icell2d' or 'no_obs',
```
dbrakenhoff, line 2065: maybe indicate that one is used for unstructured models, and the other for structured?

dbrakenhoff (commented):
Cool! I really like the idea of having a DataArray with icell2d and layer indices as values that you can quickly use to pull out heads from a DataArray at the appropriate positions! For structured I guess you might need to store x and y as well in order to be able to do something like that. 

Could we add an example of how that would work?

nlmod/dims/layers.py
OnnoEbbens: should be shape(nobs, 2)

nlmod/dims/layers.py
```diff
@@ -2003,3 +2005,116 @@ def get_isosurface(da, z, value, input_core_dims=None, exclude_dims=None, **kwar
...
+    # get model layers for screen top and bottom
+    ds_obs['modellayer_top'] = (dimname,), [np.argmax(ds_obs['screen_top'].values[i]>ds_obs['botm'].values[:,i]) for i in range(ds_obs.sizes[dimname])]
+    ds_obs['modellayer_top'] = xr.where(ds_obs['screen_top'] >= ds_obs['top'], np.inf, ds_obs['modellayer_top'])
+    ds_obs['modellayer_top'] = xr.where(ds_obs['screen_top'] <= ds_obs['botm'].isel(layer=-1), -np.inf, ds_obs['modellayer_top'])
+    ds_obs['modellayer_bot'] = (dimname,), [np.argmax(ds_obs['screen_bot'].values[i]>ds_obs['botm'].values[:,i]) for i in range(ds_obs.sizes[dimname])]
+    ds_obs['modellayer_bot'] = xr.where(ds_obs['screen_bot'] >= ds_obs['top'], np.inf, ds_obs['modellayer_bot'])
+    ds_obs['modellayer_bot'] = xr.where(ds_obs['screen_bot'] <= ds_obs['botm'].isel(layer=-1), -np.inf, ds_obs['modellayer_bot'])
+
+    # screen top is above model top but screen bottom is below model top
+    mask = (ds_obs['modellayer_top']==np.inf) & ~(ds_obs['modellayer_bot']==np.inf)
```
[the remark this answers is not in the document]
OnnoEbbens: I don't think that will work because I don't want to mask negative inf values or nan values

tests/test_009_layers.py
OnnoEbbens: Yes, also the DatasetCrossSection is tested in the 009_layers module which is a bit odd

nlmod/dims/layers.py
```diff
@@ -2003,3 +2005,116 @@ def get_isosurface(da, z, value, input_core_dims=None, exclude_dims=None, **kwar
...
+            gi = flopy.utils.GridIntersect(grid.modelgrid_from_ds(ds), method="vertex")
+            icell2d = [grid.get_icell2d_from_xy(x, y, ds, gi=gi) for x,y in xy]
+        # make dataset of observations
+        ds_obs = ds.sel(icell2d=icell2d)
+        ds_obs['screen_top'] = (('icell2d'), screen_top)
+        ds_obs['screen_bot'] = (('icell2d'), screen_bottom)
+        dimname = 'icell2d'
+    elif grid.is_structured(ds):
+        # make dataset of observations
+        dimname = 'no_obs'
```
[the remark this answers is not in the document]
OnnoEbbens: I will go for `n_obs`. I wouldn't want to use `icell2d` for structured grids because I think that is confusing.

It is only an internal function so I think `n_obs` is descriptive enough.

OnnoEbbens:
Thanks for all the enthusiasm. 

@martinvonk: nice suggestion, it will be nice to have multiple options for assigning a model layer to a well screen. I won't have time to add this now. Feel free to add this to the PR (or later with another PR).

@dbrakenhoff: interesting option. Do I understand correctly that this method will look at the screen length in relation to the thickness of a model layer? So you will need the thickness of the layer after this to obtain the same model layer that is obtained with my method?

@dbrakenhoff: In terms of a short example, I am working on adding some code to an example notebook. For now I am using it with the code below

```
oc = nlmod.read.bro.get_bro(extent, cachedir=datadir, keep_all_obs=False)
oc['modellayer'] = nlmod.layers.get_modellayers_screens(ds, oc['screen_top'], oc['screen_bottom'], xy=oc[['x','y']])
# make plots to compare model and measurements
for i, row in oc.iterrows():
    f, ax = plt.subplots()
    row['obs']['values'].plot(ax=ax, label='measurement')
    ds['heads'].sel(x=x,y=y,layer=ds.layer[int(row['modellayer'])], method='nearest').to_pandas().plot(ax=ax, label='model')
    ax.legend()
    ax.set_title(i)
```

OnnoEbbens:
So you create a dataset to use as an indexer for a model dataset. Looks neat! I will check it out.

dbrakenhoff:
Ok, sorry @OnnoEbbens, I went a bit overboard in trying to get my idea to work using your methods as the basis :sweat: Hope you don't mind the addition to your PR :innocent:.

So the idea behind this was to add two things:
1. Add get modellayers indexer, given some dataframe with observation well metadata (x, y, screen_top, screen_bottom).

```python
idx = nlmod.layers.get_modellayers_indexer(ds, df)
```
which if all goes well, should allow you to do this:
```python
h_obs = head.sel(**idx)
```

2. Interpolation of heads to locations of observation wells
```python
idx_full  = nlmod.layers.get_modellayers_indexer(ds, df, full_output=True)
hi = nlmod.observations.interpolate_points_ds(head, idx_full)
```

The first part is just a thin wrapper around your work @OnnoEbbens, but returning the result as a Dataset. The second part was largely inspired by the old hydropandas methods for pre-computing weights, but now applied to an xarray dataset. 

I added the new methods in a new file `observations.py` but I can also imagine we want to put these elsewhere. Comments are welcome!

The new example should showcase the idea behind these methods.

nlmod/plot/plot.py
```diff
@@ -477,7 +477,8 @@ def map_array(
 
     # select layer
     try:
-        nlay = da["layer"].shape[0]
+        # check if layer is a dimension
+        nlay = da["layer"].shape[0] if "layer" in da.dims else -1
```
OnnoEbbens, line 481: Is the `except KeyError` still necessary?

nlmod/observations.py
```diff
@@ -0,0 +1,232 @@
...
+    elif grid.is_vertex(da):
+        assert da.ndim == 1, "DataArray must be 1D!"
+        xy = np.vstack([x, y]).T
+    else:
+        raise ValueError("da must be structured or vertex grid")
+    vertices, weights = compute_interpolation_weights(xy, xyi, d=2)
+    return interpolate_vertices_weights(da.values, vertices, weights)
+
+
+def interpolate_points_ds(
```
OnnoEbbens, line 125: Maybe a more descriptive name is `interpolate_to_points`

nlmod/observations.py
```diff
@@ -0,0 +1,232 @@
...
+    .. [2] https://stackoverflow.com/questions/20915502/
+    speedup-scipy-griddata-for-multiple-interpolations-between-two-irregular-grids
+    """
+    ret = np.einsum("nj,nj->n", np.take(values, vertices), weights)
+    ret[np.any(weights < 0, axis=1)] = fill_value
+    return ret
+
+
+def interpolate_points(da, xi, yi, x="x", y="y"):
+    """Interpolate values at locations given by x and y coordinates.
```
OnnoEbbens, line 80: When to use this one and when to use `interpolate_points_ds`?

nlmod/observations.py
```diff
@@ -0,0 +1,232 @@
+import numpy as np
+import xarray as xr
+from scipy.spatial import Delaunay
+
+from nlmod.dims import grid
+
+
+def compute_interpolation_weights(xy, uv, d=2):
+    """Calculate interpolation weights for barycentric interpolation [1]_.
```
OnnoEbbens, line 9: I think it is better to make this function and `interpolate_vertices_weights` private functions -> `_interpolate_vertices_weights` & `_compute_interpolation_weights`

nlmod/dims/layers.py
```diff
@@ -2003,3 +2006,326 @@ def get_isosurface(da, z, value, input_core_dims=None, exclude_dims=None, **kwar
...
+    obs_ds : xr.Dataset
+        Dataset with the following variables:
+
+        - x: x-coordinate of the cell in which the observation point is located
+        - y: y-coordinate of the cell in which the observation point is located
+        - layer : model layer of each observation point
+
+        Optionally returns the following variables if full_output is True:
+
+        - icell2d: cell id of the observation point
```
OnnoEbbens, line 2208: So in case of a structured grid this is a node number?

OnnoEbbens (commented):
No worries at all, I really like the implementation!

Some comments/questions in the code and some general remarks:
- The observations are saved as a pklz file now. I think this is fine for now but eventually we would need another format because loading the pickle will certainly fail at some point. Also there is no version control on this format. I will think about an export format for hydropandas that can be put onder version control, probably some csv/zip solution.
- observation.py now only contains functions to interpolate a dataarray to points. I think we can also put this in the resample.py module.

nlmod/observations.py
OnnoEbbens: or `interpolate_da_to_points`

nlmod/dims/layers.py
```diff
@@ -2003,3 +2006,326 @@ def get_isosurface(da, z, value, input_core_dims=None, exclude_dims=None, **kwar
...
+    }
+    obs_ds = (
+        df.loc[:, [x, y, screen_top, screen_bottom]]
+        .to_xarray()
+        .rename_vars(rename_dict)
+    )
+    dim = obs_ds["x"].dims[0]  # get dimension name
+
+    # gridintersect: only use to construct list of geoms and cellids
+    gi = flopy.utils.GridIntersect(
```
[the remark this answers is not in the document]
dbrakenhoff: Gridtintersect is only used to construct a list of geometries and cellids, so yes its fast :).

nlmod/dims/layers.py
dbrakenhoff: Correct.

nlmod/observations.py
dbrakenhoff: Good idea

nlmod/observations.py
dbrakenhoff: This one works for a a simple array and x, y, xi and yi passed as arrays. I see this as a simple version thats easier to use for simple tasks, for example, if all your stuff is in numpy for example.

nlmod/observations.py
dbrakenhoff: Yea, good suggestion, those are much better. I went for interpolate to points.

nlmod/plot/plot.py
dbrakenhoff: Yes, this only checks if layer is in dims, if it is not in coords, the KeyError is raised.

tests/test_009_layers.py
dbrakenhoff: I will make issues.

dbrakenhoff:
| observation.py now only contains functions to interpolate a dataarray to points. I think we can also put this in the resample.py module.

Yea this is a good point, I had my doubts about placing it a new module. But I kind of like the idea of separating this from the resample module, since this is so directly related to obtaining time series for observation wells. (Unless you can think of another use?). We could later add more logic related to observations in this module. What do you think? If others agree this should be in resample, I will put it there. 

And good point about the pickle, I will try to reduce the file size significantly and save it as an excel and see if that works okay.

OnnoEbbens:
Yes I get what you are saying. So keeping it a separate module is good to me.

OnnoEbbens:
I cannot approve this PR because I started it. Maybe you can approve and merge yourself?

dbrakenhoff (approved):

dbrakenhoff:
Test fails unrelated.

PR #449, opened 2025-04-23
Add separation between download and gridding

OnnoEbbens:
There has been some efforts (#47) before but there were still some functions that had no separation between downloading and gridding data. This PR is an attempt to complete this separation.


rubencalje (approved):
Looks good. I did not check each line, but see a lot of  improvements.

PR #450, opened 2025-04-24
Add tests for layer methods

rubencalje:
This PR adds some tests for previously untested layer-methods.

It also fixes a problem with `get_last_active_layer_from_idomain` (which returned only zeros before) and added `get_last_active_layer` (like `get_first_active_layer`).

This PR will raise the coverage percentage to above 70 %!

image

### DIFF
--- a/tests/test_009_layers.py
+++ b/tests/test_009_layers.py
@@ -9,14 +9,13 @@ import nlmod
 from nlmod.plot import DatasetCrossSection
 
 
-def get_regis_horstermeer():
+def get_regis_horstermeer(cachedir=None, cachename="regis_horstermeer"):
     extent = [131000, 136800, 471500, 475700]
-    cachedir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "data")
+    if cachedir is None:
+        cachedir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "data")
     if not os.path.isdir(cachedir):
         os.makedirs(cachedir)
-    regis = nlmod.read.get_regis(
-        extent, cachedir=cachedir, cachename="regis_horstermeer"
-    )
+    regis = nlmod.read.get_regis(extent, cachedir=cachedir, cachename=cachename)
     return regis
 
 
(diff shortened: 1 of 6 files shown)


tests/test_009_layers.py
```diff
@@ -159,6 +158,59 @@ def test_set_minimum_layer_thickness(plot=False):
...
+    top = regis["botm"] + nlmod.layers.calculate_thickness(regis)
+    assert (top.isel(layer=layer) > z).all()
+
+
+def test_aggregate_by_weighted_mean_to_ds():
+    regis = get_regis_horstermeer()
+    regis2 = regis.copy(deep=True)
+
+    # botm needs to have the name "bottom'
+    regis2["bottom"] = regis2["botm"]
```
dbrakenhoff, line 190: Yeah, this should be fixed in the aggregate method. Let's make it a new issue.

tests/test_009_layers.py
```diff
@@ -159,6 +158,59 @@ def test_set_minimum_layer_thickness(plot=False):
...
+    # top needs to be 3d
+    regis2["top"] = regis2["botm"] + nlmod.layers.calculate_thickness(regis2)
+    kh_new = nlmod.layers.aggregate_by_weighted_mean_to_ds(regis, regis2, "kh")
+    assert np.abs(kh_new - regis["kh"]).max() < 1e-5
+    # assert (kh_new.isnull() == regis["kh"].isnull()).all() # does not assert to True...
+
+
+def test_check_elevations_consistency():
+    regis = get_regis_horstermeer()
+    nlmod.layers.check_elevations_consistency(regis)
```
dbrakenhoff, line 200: perhaps we can add an inconsistent elevation to check method when the layer model isn't correct?

tests/test_026_grid.py
```diff
@@ -134,6 +134,48 @@ def test_fillnan_da():
...
+    bgt.geometry = bgt.centroid
+    bgt["values"] = range(len(bgt))
+
+    regis = get_regis()
+    ds = nlmod.to_model_ds(regis, model_ws=model_ws)
+    sim = nlmod.sim.sim(ds)
+    gwf = nlmod.gwf.gwf(ds, sim)
+    nlmod.gwf.dis(ds, gwf)
+
+    nlmod.grid.interpolate_gdf_to_array(bgt, gwf, field="values", method="linear")
```
dbrakenhoff, line 148: we should try to bring some more structure to the interpolation methods? There are quite a few, and I've been working on a new one, but finding them isn't as easy as I'd hoped.

dbrakenhoff (approved):
Nice work! Looking good to me, I left a few comments, some can be picked up in future work.

tests/test_026_grid.py
OnnoEbbens: This notebook gives an overview of what is available: https://nlmod.readthedocs.io/en/stable/examples/07_resampling.html

tests/test_009_layers.py
rubencalje: ok, I see you did this. Thanks!

tests/test_026_grid.py
rubencalje: They are in https://nlmod.readthedocs.io/en/stable/examples/06_gridding_vector_data.html

I split the original notebook in a gridding_vector_data-notebook and a resample-notebook a long time ago.

dbrakenhoff: Thanks for the reminders, I saw the resampling notebook but I wasn't really looking for raster ops, but somehow missed the gridding_vector data notebook. I think the new work should probably get it's own notebook, but it is essentially the opposite direction of the gridding vectors notebook (but only for points). I'll commit it now so you guys can take a look.

tests/test_009_layers.py
rubencalje: I added a check. The inconsistencies are reported by the logger, but it turns out you can use pytest to test logger output as well (at least locally on my computer).

PR #453, opened 2025-05-01
Fix KNMI data platform tests

martinvonk:
Add a custom exception for KNMI data platform tests.

I tried playing around a bit with what kind of errors we generally get. Most common was the `KeyError: 'files'` because the response contained: `{'error': 'Rate limit exceeded'}`. I now check if error in the response json and raise a custom error. In the tests this custom error is also allowed.


dbrakenhoff (approved):
Nice, good solution for now I think.

PR #461, opened 2025-05-14
add map_points_to_grid

dbrakenhoff:
- method to get cellid for each point
- use method in get_modellayers_indexer
- ruff

### DIFF
--- a/nlmod/dims/layers.py
+++ b/nlmod/dims/layers.py
@@ -273,9 +273,9 @@ def split_layers_ds(
             split_dict[lay0] = [1 / split_dict[lay0]] * split_dict[lay0]
         elif hasattr(split_dict[lay0], "__iter__"):
             # make sure the fractions add up to 1
-            assert np.isclose(
-                np.sum(split_dict[lay0]), 1
-            ), f"Fractions for splitting layer '{lay0}' do not add up to 1."
+            assert np.isclose(np.sum(split_dict[lay0]), 1), (
+                f"Fractions for splitting layer '{lay0}' do not add up to 1."
+            )
             split_dict[lay0] = split_dict[lay0] / np.sum(split_dict[lay0])
         else:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


nlmod/dims/layers.py
```diff
@@ -2269,6 +2266,50 @@ def get_max_overlap_model_layer(i):
     return modellayer
 
 
+def map_points_to_grid(x, y, ds=None, modelgrid=None):
```
rubencalje, line 2269: maybe rename to get_cellids_from_xy(), and move to grid.py?

nlmod/dims/layers.py
```diff
@@ -2269,6 +2266,50 @@ def get_max_overlap_model_layer(i):
...
+    """Map points to grid cells.
```
rubencalje, line 2270: I would add a note that this method is faster than using GridIntersect

rubencalje (approved):

nlmod/dims/layers.py
```diff
@@ -2269,6 +2266,50 @@ def get_max_overlap_model_layer(i):
...
+        series with mapping between points and cellid of grid cell in
+        which points are located.
+    """
+    if ds is None and modelgrid is None:
+        raise ValueError("Either ds or modelgrid must be provided.")
+    elif ds is not None:
+        modelgrid = grid.modelgrid_from_ds(ds)
+
+    # build geometries and cellids for grid
+    gi = flopy.utils.GridIntersect(modelgrid, method="vertex")
```
rubencalje, line 2295: and make gi an optional parameter, so users can supply their own GridIntersect (just like `nlmod.grid.get_icell2d_from_xy`)

nlmod/dims/layers.py
dbrakenhoff: Good idea.

nlmod/dims/layers.py
dbrakenhoff: Done.

nlmod/dims/layers.py
dbrakenhoff: Done.

PR #464, opened 2025-05-21
fix: update URL check in _get_ahn_ellipsis to handle None values

bdestombe:
This bug was introduced with #457


rubencalje:
Thanks! Do you have the extent in which this problem occurred?

rubencalje (approved):

bdestombe:
`extent = (99000, 103000, 500000, 505000)`

Dit kwam naar voren bij het draaien van NHFLO/models/09pwnmodel2

Issue #467, opened 2025-06-05
Error reading cache if cached dataset has no 'layer' coordinate


OnnoEbbens:
PR https://github.com/gwmod/nlmod/pull/413 introduced a check of the layer dtype. This results in an error if the input dataset has the 'layer' dimension but the cached dataset does not have this.

PR #468, opened 2025-06-05
check if 'layer' dimension is in cached dataset

OnnoEbbens:
Check if the 'layer' dimension is available in the cached dataset (and not in the dataset function argument). Solves issue #467.


dbrakenhoff (approved):
My bad 🙈. Thanks for fixing!

PR #496, opened 2025-07-22
make sure layer dtype is float or int

OnnoEbbens:


OnnoEbbens:
So the title is now super confusing because I made sure the dtype is str (or float if there are nan's).

martinvonk (approved):

PR #509, opened 2025-08-11
MAW support for grouping wells

bdestombe:
The function maw_from_df now has the additional argument group, that allows for grouping of wells. This might be of use if a single pump is used to extract water from multiple wells. These individual wells may be placed in different cells, in different layers, but have to have the same radius.

    group : str, optional
        The column in df that contains the group name for the wells. If this is not
        None, wells with the same group name are grouped together such that the rate is
        divided over the wells in the group. Note that empty strings are treated as
        unique group names, so wells with an empty string in the group column are
        treated as a separate group. The default is None, which means that each well is
        treated as a separate well.

Note that boundnames have to be the same for every well within a group, which makes sense but will be a common pitfall.

### DIFF
--- a/nlmod/gwf/wells.py
+++ b/nlmod/gwf/wells.py
@@ -203,70 +203,91 @@ def maw_from_df(
     connectiondata = []
     perioddata = []
 
-    iw = 0
-    for index, irow in tqdm(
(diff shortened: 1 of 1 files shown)


bdestombe:
The errors in the tests are unrelated

PR #511, opened 2025-08-13
Speed up for fillnan

bdestombe:


bdestombe:
#511

bdestombe:
Previously, the griddata interpolation function computed a value for all cells and not just for the cells with a missing value. This PR provides a great speed up for both structured grids and unstructured grids. The speedup for structured grids in which all cells have the same shape is even bigger.

PR #529, opened 2025-12-23
Add AHN6 support and flexible AHN data configuration

rubencalje:
Introduces support for AHN6 data, including a new `download_ahn6` method and updated configuration management via `get_configuration()`.

This PR changes the default source of AHN-data to new url's shown on this page: https://basisdata.nl/hwh-portal/download/index.html. Instead of specifying a `identifier`, users can specify `version`, `resolution`, `data_kind` and `tile_size`. To still use the old-data source, a user can specify the `identifier`, of which the default version has been set to None.

The new source can download data in map tiles of 1x1 km, which is the only resolution AHN6 is available in, by setting `tile_size="1x1km"`. Users can also download data from AHN2, AHN3, AHN4 and AHN5 in map tiles of 5x6,25 km, by setting the parameter `tile_size"5x6.25km"` in the `download_ahn`-methods, which for now is also the default.

There were also legacy ahn-methods. They failed, because the webservices are not online anymore. These methods are therefore removed.

### DIFF
--- a/nlmod/read/ahn.py
+++ b/nlmod/read/ahn.py
@@ -1123,6 +1123,43 @@ def _save_tiles_from_config_in_data(config: dict = None):
                     _download_and_save_json_file(url, fname)
 
 
+def _delete_tiles_from_config_in_data(config: dict = None):
+    """Delete the location of the map tiles of AHN from the data-directory of nlmod.
+
+    This method deletes the files downloaded with `_save_tiles_from_config_in_data` from
+    the data directory again.
+
+    Parameters
+    ----------
+    config : dict, optional
+        A dictionary with properties of the data sources of the different AHN-versions.
+        When None, the configuration is retreived from the method get_configuration().
+        The default is None.
+
+    Returns
+    -------
+    None.
+
+    """
+    if config is None:
+        config = get_configuration()
+
+    pathname = os.path.join(NLMOD_DATADIR, "ahn")
+    for version in config:
+        for data_kind in config[version]:
+            for tile_size in config[version][data_kind]:
+                url = config[version][data_kind][tile_size]
+                if isinstance(url, dict):
+                    for resolution in url:
+                        fname = os.path.join(pathname, url[resolution].split("/")[-1])
+                        if os.path.isfile(fname):
+                            os.remove(fname)
+                else:
+                    fname = os.path.join(pathname, url.split("/")[-1])
+                    if os.path.isfile(fname):
+                        os.remove(fname)
+
+
 def _download_and_save_json_file(url, fname, timeout=120.0):
     response = requests.get(url, timeout=timeout)
     response.raise_for_status()


tests/test_005_external_data.py
```diff
@@ -140,10 +154,17 @@ def test_get_ahn4():
...
 
-def test_get_ahn5():
+def test_download_ahn5():
     extent = [99500.0, 100000.0, 494500.0, 495000.0]
     da = nlmod.read.ahn.download_ahn5(extent)
+    assert isinstance(da, xr.DataArray)
+    assert not da.isnull().all(), "AHN only has nan values"
+
 
+def test_download_ahn6():
```
OnnoEbbens: This is now the slowest test taking 85.18s to complete. Maybe choose a smaller extent?

nlmod/read/ahn.py
```diff
@@ -59,23 +77,26 @@ def download_ahn(
...
+        `config`. The default is None.
 
     Returns
     -------
     ahn_da : xr.DataArray
         DataArray with the ahn variable.
     """
+    if identifier is not None:
+        ahn_da = _download_ahn_ellipsis(
+            extent, identifier=identifier, merge_tiles=merge_tiles, **kwargs
```
OnnoEbbens: I think you get an error if you pass  'version', 'data_kind', 'tile_size' and 'resolution' to the `_download_ahn_ellipsis` as kwargs.

nlmod/read/ahn.py
```diff
@@ -1287,6 +1036,355 @@ def _update_ellipsis_tiles_in_data() -> None:
     tiles.to_file(fname)
 
 
+def _save_tiles_from_config_in_data(config: dict = None):
```
OnnoEbbens: missing docstring

nlmod/read/ahn.py
```diff
@@ -1287,6 +1036,355 @@ def _update_ellipsis_tiles_in_data() -> None:
...
+        fname = os.path.join(pathname, file)
+        if os.path.isfile(fname):
+            return fname
+        else:
+            return f"{base_url}/{file}"
+
+    config["AHN2"] = {
+        "DTM": {
+            "5x6.25km": {
+                0.5: get_file("AHN2_DTM05.json"),
```
OnnoEbbens: So if I understand correctly the json file is not downloaded if it is in the nlmod datadir. How does the json file get in the nlmod datadir? Do you plan on storing the data there? Should users do this themselves?

rubencalje:
> Thanks Ruben, nice to have ahn6 available and the new API calls.
> 
> I added some comments and I have two questions:
> 
> 1. What would be the preferred way of calling the `download_ahn`, is it using an identifier or using the separate, newly introduced arguments? Is one of them faster than the other? I think it would be nice to have a preference and note that in the docstring.
> 2. I don't fully get the `get_configuration` function. There seems to be an option to used stored data instead of downloading but I don't see how users should implement this. Also it may be unnecessary to download configurations for all possible combinations if you just want one of them.
> 
> Not related to this PR but I think having all this ahn data available it would be possible to get a time series of AHN values at a certain point. This makes it possible to see how the surface level changes over time.

Thanks for the review!

1. The default does not use the `identifier` method, so the preferred way is using the newly introduced arguments.
2. The tests sometimes failed because the json-file could not be downloaded. Therefore I made the method `_save_tiles_from_config_in_data`, to save all json-files in the data-directory of nlmod. If this method has been called, the downloaded json-files in the nlmod-data-directory are used. This is a bit of a hack, and not intended for public use. The 1x1 km files are quite large (16 MB), so it is not an option to just put them in the nlmod-repository.

I will try to improve the docstrings so this is clearer. Maybe we can discuss later how we can improve this workflow.

nlmod/read/ahn.py
rubencalje: The method `_save_tiles_from_config_in_data` downloads and saves all json-files to the nlmod-data-directory. This is a bit of a hack, see my comment below.

nlmod/read/ahn.py
rubencalje: Implemented a warning and removal from kwargs

nlmod/read/ahn.py
```diff
@@ -1275,7 +1014,17 @@ def download_ahn5(
     xr.DataArray
         DataArray of the AHN
     """
-    return _download_ahn_ellipsis(extent, identifier, **kwargs)
+    if identifier is not None:
+        return _download_ahn_ellipsis(extent, identifier, **kwargs)
+    return _download_ahn_hwh(extent=extent, version="AHN5", **kwargs)
+
+
+def download_ahn6(extent: list[float], tile_size: str = "1x1km", **kwargs):
```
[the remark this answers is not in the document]
rubencalje: Added docstring

nlmod/read/ahn.py
```diff
@@ -59,23 +81,31 @@ def download_ahn(
...
 
     Returns
     -------
     ahn_da : xr.DataArray
         DataArray with the ahn variable.
     """
+    if identifier is not None:
+        for key in ["version", "data_kind", "tile_size", "resolution"]:
+            if key in kwargs:
+                logger.warning("`{key}` is ignored then `identifier` is sepcified")
```
OnnoEbbens:
```suggestion
                logger.warning("`{key}` is ignored when `identifier` is specified")
```

PR #552, opened 2026-03-11
Remove ecmwflibs from dependencies

rubencalje:
ecmwflibs is not available for python 3.12 or newer, therefore it may be better to remove the dependency


martinvonk:
Good idea probably. Related issue from https://github.com/ecmwf/ecmwflibs/issues/33 which mentions `eccodes` as an alternative but I think we barely use this anyway.

PR #553, opened 2026-03-17
Improve tests and notebooks

rubencalje:
This PR makes sure notebooks and tests produce output in folder that are not committed to Git:
- the models in the notebooks are run in a folder with the same name as the notebook. we make sure this folder is included in gitignore.
- the tests are run inside a temporary folder that is written to an environmental variable, so it can be used by other tests. This environmental variable is created when test are created outside pytest.

Also this PR improves some tests:
- remove 3 of the sea_model tests in test_001_model.py, as these tests only add 3 extra covered lines, have a lot of duplicate code and take a long time.
- Use `tmp_path` instead of `tmpdir`
- Fix some of the warnings pytest shows
- Make sure to close all figures and close handles to netcdf file at every test (in conftest.py). This might have been the reason for the freezed tests before.
- Add extra tests using copilot
- add test for `nlmod.read.administrative.download_kadaster_percelen`
- do not use deprecated methods in anymore, but use the replacements (`download_` instead of `get_`)

Some other changes:
- remove broken 'year' argument in `download_percelen_gdf`
- remove tight_layout where applicable from plot-methods
- add custom.css so dataframes in documenttaion look nice again (after the update of myst-nb)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,10 @@ jobs:
         python-version: [3.11]
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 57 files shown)


dbrakenhoff (approved):
Looks good. Didn´t check it all super carefully, and the new monkeypatch tests are a bit hard to follow but I'm sure if the tests pass and run quickly this is is a big improvement.

rubencalje:
Yes, the monkeypatch-stuff is what copilot came up with when I asked to increase the coverage. They are all in the last two test-files, so easy to remove again.

PR #556, opened 2026-03-27
add new methods for geulen

OnnoEbbens:
Add two new methods for merging geotop geulen into a geotop stratigraphic model.


OnnoEbbens:
This PR is ready for review. Although I think it will be pretty hard to review.

I tried to take into account all possible scenarios for this tool, but still I added a warning because I am not sure the method works for all edge cases (there are so many!).

There are a few assumptions in the code:
- A geul unit with a higher number (e.g. 6420) is always below a geul unit with a lower number (e.g. 6010).
- A geul unit is never interrupted by another unit in the vertical. Meaning there is never another unit in between geul units when looking at a column. This does not hold the other way around, a geul unit can be both above and below another unit.

In the PR there are now 2 functions `split_layers_on_geul` and `split_layers_on_geul_optimal`. As the name suggest I think the last one is better as it is slightly faster and the final layer model has less layers. However the code is definitely more complex and more steps are involved. I mainly left both methods in so we can easily go back to that method if there are problems with the 'optimal' one.

The 15_geotop.ipynb has code to show the result from both methods. This might be the easiest way to check the results. I will remove one of the methods and clean up the notebook once we choose.

OnnoEbbens:
@dbrakenhoff and @rubencalje, can you wait with the review? Copilot has some good points that I want to solve first.

OnnoEbbens:
@rubencalje & @dbrakenhoff this PR is finally ready for review. Failing test not related to this PR.

The PR adds the function `split_layers_on_geul` which makes it possible to add geulen as separate layers instead of adding geulen to the layer above or the layer below. This comes at the cost of adding extra layers to the layer model. The idea is easiest to understand by having a look at the cross section below. In that example 4 extra layers are added using the `split_layers_on_geul` function.

Note: there was already an option "add_as_layer" in the `nlmod.read.geotop.to_model_layers` but that raised an error if geulen are both above and below other stratigraphic layers. I think we can keep that option for the cases where a geul can be interpreted as a "normal" stratigraphic layer.

```
units in original gt: 16
layers in gtl1: 14
layers in gtl2: 14
layers in gtl3: 20
```

image

nlmod/read/geotop.py
```diff
@@ -75,6 +75,258 @@ def get_kh_kv_table(kind="Brabant"):
     return df
 
 
+def split_layers_on_geul(strat, units_no_geul, geulen):
```
dbrakenhoff, line 78: I still find the dutch "geul" a bit awkward, but I think for most users this is still clearer than using paleochannels or somethig like that, so "geulen" it is :)

nlmod/read/geotop.py
```diff
@@ -75,6 +75,258 @@ def get_kh_kv_table(kind="Brabant"):
...
+        )  # index (+1) of unit directly below the geul
+        bot_lay_geul[
+            np.isnan(bot_geul)
+        ] = -999  # -999 if geul is not present in vertical
+        bot_lay_geul[bot_lay_geul == 0] = -888  # -888 if layer below geul is also geul.
+        bot_lay_geul[bot_geul == z_bot] = 999  # 999 if geul is the bottom layer
+
+        bot_lay_geul = np.where(
+            bot_lay_geul == -888, (bot_geul > bot).argmax(axis=0), bot_lay_geul
+        )  # if layer below geul is also geul, take the layer below that geul :)
```
dbrakenhoff: What would happen if there are three stacked "geulen"?

nlmod/read/geotop.py
```diff
@@ -75,6 +75,258 @@ def get_kh_kv_table(kind="Brabant"):
...
+                strat = xr.where(mask4 & mask1, geul_subset, strat)
+                new_unit_order = (
+                    new_unit_order[: ilay + 1]
+                    + [geul_subset]
+                    + new_unit_order[ilay + 1 :]
+                )
+        logger.debug(f"new order of units: {new_unit_order}")
+        unit_order = new_unit_order.copy()
+
+    return strat, new_unit_order
```
dbrakenhoff, line 342: Impressive code 🤓 ! And I appreciate the comments everywhere :).

nlmod/read/geotop.py
OnnoEbbens: That would also work, comment should be `# if layer below geul is also geul take the first non-geul layer.`
```suggestion
        )  # if layer below geul is also geul, take the first non-geul layer.
```

nlmod/read/geotop.py
OnnoEbbens: Yes it is a bit awkward but also conveniently short

nlmod/read/geotop.py
OnnoEbbens: Thanks!

dbrakenhoff:
Really nice work! Approved, but I did run into a small bug maybe, so maybe needs one more check? I wrote some tiny code to test a few examples. It produces these kinds of figures (perhaps the basis of some testing...).

image

I haven't exhausted all the possible, examples so we probably still need to add a few. But I saw that adding a geul at the bottom is currently causing an index error. Perhaps you can check what's going on for that edge case?

```python
# %%
import matplotlib as mpl
import matplotlib.pyplot as plt
import numpy as np
import xarray as xr

from nlmod.read.geotop import split_layers_on_geul


def plot_stratigraphy(strat_original, strat_modified):

    unique_values = np.unique(np.concatenate([strat_original, strat_modified]))
    boundaries = np.append(unique_values, unique_values[-1] + 1)
    norm = mpl.colors.BoundaryNorm(boundaries, len(unique_values))
    cmap = plt.get_cmap("Dark2", len(unique_values))

    fig, (axl, axr) = plt.subplots(1, 2, sharex=True, sharey=True)
    axl.matshow(strat.squeeze(), norm=norm, cmap=cmap)
    axr.matshow(strat_modified.squeeze(), norm=norm, cmap=cmap)
    axl.set_title("Original Stratigraphy")
    axr.set_title("Modified Stratigraphy")
    # Loop through the array to add text
    for istrat, iax in zip([strat_original, strat_modified], (axl, axr)):
        rows, _, cols = strat_original.shape
        for i in range(rows):
            for j in range(cols):
                val = istrat[i, 0, j]
                # Use white text for dark colors, black for light colors
                color = "white"
                iax.text(
                    j,
                    i,
                    str(val.item()),
                    va="center",
                    ha="center",
                    color=color,
                    fontweight="bold",
                )
    return


# %% example split single layer
strat = xr.DataArray(
    np.array([[0, 0, 0], [0, 10, 10], [0, 0, 0]])[:, None, :],
    coords={"z": [0, -1, -2], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)

# %%
# %% example geul at top
strat = xr.DataArray(
    np.array([[0, 10, 10], [1, 1, 1], [2, 2, 2]])[:, None, :],
    coords={"z": [0, -1, -2], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1, 2]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)

# %% example geul at bottom
strat = xr.DataArray(
    np.array([[0, 0, 0], [1, 1, 1], [2, 10, 10]])[:, None, :],
    coords={"z": [0, -1, -2], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1, 2]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)

# %% example simple add between layers
strat = xr.DataArray(
    np.array([[0, 0, 0], [0, 10, 10], [1, 1, 1]])[:, None, :],
    coords={"z": [0, -1, -2], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)

# %% example add between layers with geul at top and bottom
strat = xr.DataArray(
    np.array([[0, 0, 10], [10, 10, 10], [10, 1, 1]])[:, None, :],
    coords={"z": [0, -1, -2], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)

# %% example 2 geulen
strat = xr.DataArray(
    np.array([[0, 10, 0], [1, 20, 1], [2, 2, 2]])[:, None, :],
    coords={"z": [0, -1, -2], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1, 2]
geulen = [10, 20]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
```

dbrakenhoff (approved):

OnnoEbbens:
Thanks a lot! I was struggling with the tests myself, your approach is very clean. I added the tests and made some improvements for the edge cases.

I expanded your script and pasted it below in case we ever need to debug this.

```python

import matplotlib as mpl
import matplotlib.pyplot as plt
import numpy as np
import xarray as xr

from nlmod.read.geotop import split_layers_on_geul


def plot_stratigraphy(strat_original, strat_modified):

    unique_values = np.unique(np.concatenate([strat_original, strat_modified]))
    boundaries = np.append(unique_values, unique_values[-1] + 1)
    norm = mpl.colors.BoundaryNorm(boundaries, len(unique_values))
    cmap = plt.get_cmap("Dark2", len(unique_values))

    fig, (axl, axr) = plt.subplots(1, 2, sharex=True, sharey=True)
    axl.matshow(strat.squeeze(), norm=norm, cmap=cmap)
    axr.matshow(strat_modified.squeeze(), norm=norm, cmap=cmap)
    axl.set_title("Original Stratigraphy")
    axr.set_title("Modified Stratigraphy")
    # Loop through the array to add text
    for istrat, iax in zip([strat_original, strat_modified], (axl, axr)):
        rows, _, cols = strat_original.shape
        for i in range(rows):
            for j in range(cols):
                val = istrat[i, 0, j]
                # Use white text for dark colors, black for light colors
                color = "white"
                iax.text(
                    j,
                    i,
                    str(val.item()),
                    va="center",
                    ha="center",
                    color=color,
                    fontweight="bold",
                )
    return

# %% example 2 geulen
strat = xr.DataArray(
    np.array([[0, 10, 0], [1, 20, 1], [2, 2, 2]])[:, None, :],
    coords={"z": [0, -0.5, -1.0], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1, 2]
geulen = [10, 20]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)

# %% example geul at top
strat = xr.DataArray(
    np.array([[0, 10, 10], [1, 1, 1], [2, 2, 2]])[:, None, :],
    coords={"z": [0, -0.5, -1.0], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1, 2]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)

# %% example geul between and below
strat = xr.DataArray(
    np.array([[0, 0, 0], [10, 1, 1], [0,10, 10]])[:, None, :],
    coords={"z": [0, -0.5, -1.0], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)

# %% example split single layer
strat = xr.DataArray(
    np.array([[0, 0, 0], [0, 10, 10], [0, 0, 0]])[:, None, :],
    coords={"z": [0, -0.5, -1.0], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)

# %% example geul at bottom
strat = xr.DataArray(
    np.array([[0, 0, 0], [1, 1, 1], [2, 10, 10]])[:, None, :],
    coords={"z": [0, -0.5, -1.0], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1, 2]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)

# %% example simple add between layers
strat = xr.DataArray(
    np.array([[0, 0, 0], [0, 10, 10], [1, 1, 1]])[:, None, :],
    coords={"z": [0, -0.5, -1.0], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)

# %% example add between layers with geul at top and bottom
strat = xr.DataArray(
    np.array([[0, 0, 10], [10, 10, 10], [10, 1, 1]])[:, None, :],
    coords={"z": [0, -0.5, -1.0], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)

# %% example geul between and below
strat = xr.DataArray(
    np.array([[0, 0, 0], [10, 1, 1], [0,10, 10], [0,2, 2]])[:, None, :],
    coords={"z": [0, -0.5, -1.0, -1.5], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1,2]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)

# %% example geul between and below
strat = xr.DataArray(
    np.array([[0, 0, 0], [10, 1, 1], [0,10, 10]])[:, None, :],
    coords={"z": [0, -0.5, -1.0], "y": [100], "x": [100, 200, 300]},
    dims=("z", "y", "x"),
)
units_no_geul = [0, 1]
geulen = [10]

strat_modified, new_unit_order = split_layers_on_geul(strat, units_no_geul, geulen)
plot_stratigraphy(strat, strat_modified)
print(new_unit_order)
```

dbrakenhoff:
Alright, looks good! Just a quick final check:

image

This new layer order is what we would expect right? And I guess it doesn't really matter since 0 and 1 never overlap here? Because it does look a bit odd that 0 is now placed earlier in the new order than 1.

OnnoEbbens:
Yes, the order is correct and the naming is definitely odd.

It is not easy to change this in the `split_layers_on_geul` function. I think the easiest would be to change this during the conversion of the unit numbers to stratigraphic names.

OnnoEbbens:
Fixed it in the `to_model_layer` function so the layer order in the end result makes more sense.

PR #562, opened 2026-07-01
Improve lake gwt

rubencalje:
This PR improves the support for a GroundWater Transport (GWT) model in the Lake (LAK) package. The user can now supply the concentration of the fluxes in and out of the lake, in the same way he can supply the fluxes for the GroundWater Flow (GWF) model.

### DIFF
--- a/nlmod/gwf/lake.py
+++ b/nlmod/gwf/lake.py
@@ -24,6 +24,15 @@ LAKE_KWDS = [
     "ROUGH",
 ]
 
+LAKE_GWT_KWDS = [
+    "GWT_STATUS",
+    "GWT_CONCENTRATION",
+    "GWT_RAINFALL",
+    "GWT_EVAPORATION",
+    "GWT_RUNOFF",
+    "GWT_EXT-INFLOW",
+]
+
 # order of dictionary matters!
 OUTLET_DEFAULT = {
     "couttype": "WEIR",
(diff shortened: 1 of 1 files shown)


nlmod/gwf/lake.py
```diff
@@ -207,6 +222,13 @@ def lake_from_gdf(
             and not lake_gdf["lakeout"].eq("").all()
         ):
             lakeout = _get_and_check_single_value(lake_gdf, "lakeout")
+            if isinstance(lakeout, str):
```
bdestombe, line 225: Shouldn't this logic be part of _get_and_check_single_value()?

bdestombe: Maybe it is way easier and robust to just expect boundnames, and to be strings? I don't like this behavior of using a single argument for either indices or names. The indices seem fragile anyway, when adding a feature, suddenly the numbering has to be checked

rubencalje: Only lakeout can be an integer as well. If we put this check in `_get_and_check_single_value`, unexpected things can happen to other variables.

I agree it is a bit ugly, that lakeout can be an integer (the lake number) or a string (the boundname). But, for backwards compatibility, I would like to keep both.

dbrakenhoff (approved):
Looks good!

PR #563, opened 2026-07-05
added 'na_filter=False' to read_csv in read_gleg in nlmod.read.regis.py

vincentpost:
Suggest to add na_filter=False to pd.read_csv() in read_gleg. Otherwise the code for the "Formatie van Naaldwijk", which is "NA",  mistakenly gets recognized as a nan value.


rubencalje:
Great, thank you!

This sounds familiar, maybe the reason for some strange plots in the past.

rubencalje (approved):

PR #565, opened 2026-07-08
Add HFB package builder from dataframes

bdestombe:
The full workflow from GeoDataFrame line features to a FloPy HFB package is implemented in nlmod, while delegating the horizontal line-to-cell routing to FloPy.

## Summary
- add `nlmod.gwf.hfb.hfb_from_df` to build MF6 HFB packages from line GeoDataFrames
- delegate HFB line routing to `flopy.utils.make_hfb_array` from modflowpy/flopy#2745
- keep nlmod-specific depth/elevation `hydchr` scaling and `idomain <= 0` filtering around the FloPy-generated cell pairs
- use FloPy `develop` until the HFB utility from flopy#2745 is available in a release (`flopy>3.10.0`)
- cover default columns, scalar values, custom columns, MultiLineString input, package naming, zero-SPD cleanup, FloPy routing, and inactive/pass-through filtering in HFB tests

## Validation
- `python -m py_compile nlmod\gwf\hfb.py`
- `python -m ruff format nlmod\gwf\hfb.py tests\test_023_hfb.py`
- `python -m ruff check --fix nlmod\gwf\hfb.py tests\test_023_hfb.py`
- `python -m pytest tests\test_023_hfb.py -v`
- `python -m prospector nlmod\gwf\hfb.py tests\test_023_hfb.py`
- `bandit -q -r nlmod\gwf\hfb.py`
- local availability check for `flopy.utils.make_hfb_array`

### DIFF
--- a/nlmod/gwf/hfb.py
+++ b/nlmod/gwf/hfb.py
@@ -29,6 +29,7 @@ def hfb_from_df(
     df,
     gwf,
     ds,
+    *,
     hydchr="hydchr",
     depth="depth",
     elevation="elevation",
@@ -80,7 +81,7 @@ def hfb_from_df(
     if unsupported:
         raise ValueError(
             "hfb_from_df supports only LineString and MultiLineString geometries; "
-            f"found {sorted(unsupported)}"
+            f"found {sorted(unsupported)}",
         )
 
     spd = []
@@ -90,7 +91,7 @@ def hfb_from_df(
             raise ValueError(f"hydchr must be set for HFB feature at index {index!r}")
         if hydchr_value <= 0:
             raise ValueError(
-                f"hydchr must be positive for HFB feature at index {index!r}"
+                f"hydchr must be positive for HFB feature at index {index!r}",
             )
 
         depth_value = _get_optional_hfb_value(row, depth, "depth", index)
@@ -98,7 +99,7 @@ def hfb_from_df(
         if (depth_value is None) == (elevation_value is None):
             raise ValueError(
                 "Exactly one of depth or elevation must be set for HFB feature "
-                f"at index {index!r}"
+                f"at index {index!r}",
             )
 
         feature = df.iloc[[row_number]]
@@ -106,7 +107,10 @@ def hfb_from_df(
             spd += get_hfb_spd(ds, feature, hydchr=hydchr_value, depth=depth_value)
         else:
             spd += get_hfb_spd(
-                ds, feature, hydchr=hydchr_value, elevation=elevation_value
+                ds,
+                feature,
+                hydchr=hydchr_value,
+                elevation=elevation_value,
             )
 
     spd = _clean_hfb_spd(spd)
@@ -131,7 +135,7 @@ def _get_optional_hfb_value(row, value_or_column, name, index):
         if value_or_column not in row.index:
             raise KeyError(
                 f"Column {value_or_column!r} not found for {name} of HFB feature "
-                f"at index {index!r}"
+                f"at index {index!r}",
             )
         value = row[value_or_column]
     else:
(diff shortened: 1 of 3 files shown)


nlmod/gwf/hfb.py
```diff
@@ -371,7 +510,7 @@ def line_to_hfb_buffer(gdf, ds, buffer_distance=None, gi=None):
     if buffer_distance is None:
         # buffer distance ~ 1.5 * max cell size
         buffer_distance = 1.5 * np.max(
-            [np.max(np.abs(np.diff(ds["x"]))), np.max(np.abs(np.diff(ds["y"])))]
+            [np.max(np.abs(np.diff(ds["x"]))), np.max(np.abs(np.diff(ds["y"])))],
```
dbrakenhoff, line 529: What's with the trailing commas?

dbrakenhoff (approved):
Looks good to me!

Seeing this reminds me that flopy now has HFB logic: https://github.com/modflowpy/flopy/pull/2745

So I'm not sure how robust our algorithm is, but it could be worth exploring deferring the logic for building the HFBs to flopy? And on a similar note, I at some point added the `line_to_hfb_buffer` method that was supposed to be more robust and faster for more complex HFBs. I don't know if I actually succeeded, but it could be worth checking out this method as well... 

But feel free to open a new issue to tackle those questions :)

bdestombe:
> Looks good to me!
> 
> Seeing this reminds me that flopy now has HFB logic: https://github.com/modflowpy/flopy/pull/2745
> 
> So I'm not sure how robust our algorithm is, but it could be worth exploring deferring the logic for building the HFBs to flopy? And on a similar note, I at some point added the `line_to_hfb_buffer` method that was supposed to be more robust and faster for more complex HFBs. I don't know if I actually succeeded, but it could be worth checking out this method as well... 
> 
> But feel free to open a new issue to tackle those questions :)

Thanks for the review! And I'm all in favor, but that would require pinning flopy to something recent. Could that cause any issues with other users?

As for the linting/formatting, I'll update that. I think I was developing from a different folder so that ruff didn't pick up on the formatting configuration file.

dbrakenhoff:
> Thanks for the review! And I'm all in favor, but that would require pinning flopy to something recent. Could that cause any issues with other users?

Good question. I don't really think so since flopy mostly manages I/O and backwards compatibility is something they take very seriously.

bdestombe:
Thanks for the pointer to modflowpy/flopy#2745. I updated this PR to use that FloPy logic instead of maintaining nlmod's own line-routing path for the new package-builder workflow.

Changes since your review:
- `get_hfb_spd` now delegates horizontal line-to-cell routing to `flopy.utils.make_hfb_array`.
- nlmod still handles the parts that FloPy does not: depth/elevation vertical `hydchr` scaling and explicit `idomain <= 0` filtering.
- `hfb_from_df` remains the GeoDataFrame-to-package convenience wrapper on top of `get_hfb_spd`.
- The FloPy dependency now points to the `develop` branch, with an inline `pyproject.toml` comment noting that this is for the HFB utility from flopy#2745 and can move to `flopy>3.10.0` once released.
- Tests now assert FloPy availability/use, structured known-answer depth/elevation scaling, vertex cell-pair agreement with direct FloPy output, MultiLineString handling, and inactive/pass-through filtering completeness.
- I also removed the stale commented-out endpoint-handling block in the legacy `line_to_hfb` path.

Local validation passed with FloPy `develop`: `py_compile`, ruff format/check, `pytest tests\test_023_hfb.py -v`, prospector, bandit, and an explicit `flopy.utils.make_hfb_array` import check.

bdestombe:
As far as I can tell there is no option to disable the trailing comma in multiline calls in the ruff configuration. Apparently, this complies with the standard, but also nlmod's, formatting rules

nlmod/gwf/hfb.py
bdestombe: As far as I can tell there is no option to disable the trailing comma in multiline calls in the ruff configuration. Apparently, this complies with the standard, but also nlmod's, formatting rules

dbrakenhoff: You're right, apparently it accepts both. I thought that after single argument functions it usually avoided a comma, but I guess that's just because I never add a comma after writing code like that. 

It looks a bit weird to me, but that's just what I've gotten used to. So fine by me :+1: !

bdestombe:
Hi David and others,
Merging this would require nlmod to use the develop branch of flopy. I would understand this is not ideal, however:
- It would only be until the next flopy version is released (e.g., flopy version > 3.10)
- it passes the current nlmod CI, so seems stable enough

dbrakenhoff:
Yeah, ideally we release this only when flopy releases with the hfb utils and pin nlmod to use the newest flopy? From issues on flopy I gather the next release should happen in the next 1-2 weeks, so we won't have to wait that long?

So fine by me to merge this into dev, and hold our next release (which we really should do) until flopy has made theirs.

PR #567, opened 2026-07-09
improve surface water

dbrakenhoff:
- ensure season timeseries is defined for last season in model
- add small eps to rbot to avoid floating point issues with mf6


rubencalje:
find out if "+ 1e-6" is really necessary

dbrakenhoff:
Removed the 1e-6 adjustments, will try to implement those in the project-specific code.

PR #568, opened 2026-07-09
Add method to build mf6 observation file from dataframe with x, y, z …

dbrakenhoff:
…positions


rubencalje (approved):
Docstring would be great

PR #571, opened 2026-07-09
add apply_tsmult helper function

dbrakenhoff:


rubencalje:
This is a duplicate to `nlmod.time.get_time_step_length` I think. Do you get the same result?

dbrakenhoff:
Same result, closing PR.

PR #576, opened 2026-07-17
Fix inverted HYDCHR scaling for partially penetrated HFB layers

bdestombe:
## Problem

`get_hfb_spd` scales `hydchr` by the penetrated fraction of the layer that contains the bottom of the barrier:

```python
hydchr_frac = (depth - layer_top_depth) / thicki[ilay]
spd.append([cellid1, cellid2, hydchr * hydchr_frac])
```

But a **lower** HYDCHR is a **stronger** barrier in MF6 (the barrier conductance is `HYDCHR × face area`, placed in series with the cell-to-cell conductance). So a wall that penetrates only 10% of a layer is modeled as a barrier ~10× *tighter* than a fully penetrating wall, instead of leaving the un-penetrated 90% of that face open. The effect is upside-down: the less a barrier penetrates a layer, the more it blocks it.

A two-cell MF6 experiment (10 m layer, K = 10 m/d, unit gradient):

| face | flow |
|------|------|
| no barrier | 100.0 m³/d |
| full wall, hydchr = 0.01 | 9.09 m³/d |
| nlmod scaling for 10% penetration (hydchr = 0.001) | 0.99 m³/d |
| physical target for 10% penetration | ≳ 90 m³/d |

## Fix

For the layer containing the barrier bottom, use a parallel-path equivalent characteristic: the open strip over the un-penetrated fraction `1 − f` in parallel with the walled strip over `f` (the latter in series with the aquifer). Equating the MF6 series form to that combination gives

```
hydchr_eff = (hydchr + (1 − f) · c) / f
```

with `c = kh_harmonic / distance` the open-face conductance per unit area. This reduces to `hydchr` at full penetration (`f = 1`) and leaves the face open (`hydchr_eff → ∞`) as `f → 0`.

## Validation

New test `test_get_hfb_spd_partial_penetration_matches_resolved_barrier` compares the equivalent single-layer barrier against a reference MF6 run that resolves the partially penetrated layer into a walled and an open sublayer — the flows match within 1%. The previous scaling underestimates that reference flow by more than a factor of 10. The existing structured/vertex/multilinestring expectations were updated to the parallel-equivalent values. Full `tests/test_023_hfb.py` passes (23 tests).

Surfaced while reviewing the Bergen sheet-pile walls in the NHFLO `09pwnmodel2` model, which drive `hfb_from_df` → `get_hfb_spd` with `depth` barriers.

### DIFF
--- a/nlmod/gwf/hfb.py
+++ b/nlmod/gwf/hfb.py
@@ -195,6 +195,11 @@ def get_hfb_spd(ds, linestrings, hydchr, depth=None, elevation=None):
     thickness at the cell interface is just the average of the thicknesses of the two
     cells.
 
+    For the layer that contains the bottom of the barrier, an equivalent hydraulic
+    characteristic is used that combines the open part of the face (below the barrier)
+    in parallel with the walled part, so a barrier that barely penetrates a layer
+    leaves that face effectively open instead of blocking it.
+
     Parameters
     ----------
     ds : xr.Dataset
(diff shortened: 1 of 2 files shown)


dbrakenhoff (approved):
Good fix! As a side note, I think modflow also accepts negative `hydchr` in which case it is interpreted as a factor on the cell-to-cell conductance, which could be useful in the context of a partially penetrating impermeable walls. Not sure if we want to deal with that here, but just mentioning it in case it is of value.

PR #581, opened 2026-07-21
add gwfgwf exchange func

dbrakenhoff:
- add gwfgwf exhange package
- add method to compute exchange data using geopandas from parent and child model datasets.


nlmod/sim/sim.py
```diff
@@ -259,3 +261,154 @@ def register_ims_package(sim, model, ims):
 
 def register_solution_package(sim, model, solver):
     sim.register_solution_package(solver, [model.name])
+
+
+def get_parent_child_exchange_gdf(ds_parent, ds_child, boundnames="angldegx"):
```
OnnoEbbens, line 266: Can you  make a docstring?

nlmod/sim/sim.py
```diff
@@ -259,3 +261,154 @@ def register_ims_package(sim, model, ims):
...
+    gdf_parent = modelgrid_from_ds(ds_parent).to_geodataframe()
```
OnnoEbbens, line 287: Just out of curiosity. We now use 2 methods to get a geodataframe with the modelgrid. `modelgrid_from_ds(ds_parent).to_geodataframe()` & `nlmod.grid.polygons_from_ds`. Any idea which one should be the preferred one?

OnnoEbbens (changes requested):
Just some minor remarks. Also I did not check everything in detail because that would take too much time. How sure are you these functions will work for all cases (rotated grids, vertex/structured grids)? If not I would suggest to add a slight warning that this is still an experimental feature. I leave that up to you

nlmod/sim/sim.py
dbrakenhoff: `polygons_from_ds()` returns only the polygons, so you'd still have to build the GeoDataFrame from that. It also internally calls `modelgrid_from_ds()` for vertex grids, so I say, let's defer to flopy where we can :).

dbrakenhoff:
>  If not I would suggest to add a slight warning that this is still an experimental feature. I leave that up to you

It should work for vertex and structured grids, but there is the requirement that the grids have to align exactly. As long as the grids are built in a similar way, I think it will work fine, but it's not been thoroughly tested. I think it's okay to add it without warnings though.

OnnoEbbens (approved):

PR #582, opened 2026-07-22
Aggregate lake pieces per cell in lake_from_gdf, like RIV/DRN celldata aggregation

bdestombe:
## What

`lake_from_gdf` now accepts per-piece input — multiple rows per cellid for one lake, as produced directly by `nlmod.grid.gdf_to_grid` — and combines the pieces into one lake-GWF connection per cell with an area-weighted bed resistance. It also tolerates float-level noise in numeric per-lake settings.

## Why

Two defects when the input is not pre-aggregated by the caller:

1. **Per-piece rows multiply-count the lake-aquifer exchange.** MODFLOW 6 applies each VERTICAL lake-GWF connection over the **full cell area**, and `lake_from_gdf` turns every row into its own connection with `bedleak = 1/clake`. A cell holding two pieces of one lake (6000 m² at `clake=10` and 2000 m² at `clake=20` in a 10000 m² cell) gets two full-cell-area connections totalling `0.15 * cell_area` of conductance, where the pieces physically justify `(6000/10 + 2000/20) / 10000 = 0.07 * cell_area`.
2. **Exact-equality single-value check rejects aggregated inputs.** `_get_and_check_single_value` compares `strt` (and outlet settings) with `==`. Area-weighted per-cell stages carry float-level noise (a 1e-12 difference), which raised `AssertionError: A single lake should have a single strt` for no physical reason.

RIV and DRN already have exactly this aggregation step in nlmod: `nlmod.gwf.surface_water.aggregate` collapses per-piece celldata to one reach per cell with area-weighted parameters before `build_spd`. The LAK path lacked the equivalent, forcing every caller to hand-roll it. (Encountered wiring the NHFLO Bergen ponds — NHFLO/models#127, NHFLO/tools#58 — where the workaround currently lives downstream.)

## How

- New `_aggregate_connections_per_cell(lake_gdf, ds)`, invoked per lake when the cellid index has duplicates: combined `clake = cell_area / sum(piece_area / clake)`, so the VERTICAL connection conductance `cell_area / clake` equals the summed piece conductance; other columns take the first piece's value (they are per-lake settings, still validated by `_get_and_check_single_value`). Raises a clear `ValueError` when pieces come without geometry, since the areas are then unknown.
- `_get_and_check_single_value` uses `np.allclose(rtol=1e-8, atol=0)` for numeric columns; exact comparison is kept for strings (boundnames, timeseries names, `lakeout`).
- `nlakeconn` consequently counts cells, not pieces.

Behaviour for already-aggregated input (one row per cell per lake) is unchanged.

## Tests

Two regression tests in `test_013_surface_water.py`, both shown to fail on unchanged `dev` before the fix:

- `test_lake_from_gdf_aggregates_pieces_per_cell` — fails on dev with 3 connections at `bedleak=0.1`/`0.05` instead of 2 connections at the area-weighted `0.07`/`0.1`; also pins `nlakeconn == 2`.
- `test_lake_from_gdf_accepts_floating_point_strt_noise` — fails on dev with the `AssertionError` on a `1e-12` strt difference.

`test_gdf_lake` (existing) still passes; `ruff check`/`ruff format` add no findings relative to dev.

### DIFF
--- a/nlmod/gwf/lake.py
+++ b/nlmod/gwf/lake.py
@@ -62,6 +62,11 @@ def lake_from_gdf(
             lakeno : with the number of the lake
             strt : with the starting head of the lake
             clake : with the bed resistance of the lake
+        A lake may have multiple rows (polygon pieces) per cellid, e.g. straight
+        from nlmod.grid.gdf_to_grid. Pieces are combined into one lake-GWF
+        connection per cell with an area-weighted bed resistance (see
+        _aggregate_connections_per_cell), like nlmod.gwf.surface_water.aggregate
+        already does for RIV/DRN celldata.
             optional columns are 'STATUS', 'STAGE', 'RAINFALL', 'EVAPORATION',
             'RUNOFF', 'INFLOW', 'WITHDRAWAL', 'AUXILIARY', 'RATE', 'INVERT',
             'WIDTH', 'SLOPE', 'ROUGH'. These columns should contain the name
@@ -158,6 +163,8 @@ def lake_from_gdf(
         gdf = add_lakeno_to_gdf(gdf, boundname_column)
 
     for lakeno, lake_gdf in gdf.groupby("lakeno"):
+        if lake_gdf.index.duplicated().any():
+            lake_gdf = _aggregate_connections_per_cell(lake_gdf, ds)
         nlakeconn = lake_gdf.shape[0]
         if "strt" in lake_gdf:
             strt = _get_and_check_single_value(lake_gdf, "strt")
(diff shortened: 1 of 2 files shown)


nlmod/gwf/lake.py
```diff
@@ -349,10 +356,39 @@ def lake_from_gdf(
...
+    connection over the full cell area, so per-piece rows would multiply-count the
+    exchange. Each piece therefore contributes conductance for its own area: the
+    combined bed resistance is clake = cell_area / sum(piece_area / clake), so
+    that the connection conductance cell_area / clake equals the summed piece
+    conductance — the same area-weighted aggregation nlmod.gwf.surface_water
+    .aggregate applies to RIV/DRN celldata. All other columns take the value of
+    the first piece per cell.
+    """
+    if "geometry" not in lake_gdf.columns or lake_gdf.geometry.isna().any():
+        raise ValueError(
```
OnnoEbbens, line 373: I had a bit of trouble reading the error message. Is this a good alternative:

 "found multiple elements of one lake in a single cell; provide polygon "
 "geometries of the lake to aggregate these elements"

nlmod/gwf/lake.py
```diff
@@ -62,6 +62,11 @@ def lake_from_gdf(
             lakeno : with the number of the lake
             strt : with the starting head of the lake
             clake : with the bed resistance of the lake
+        A lake may have multiple rows (polygon pieces) per cellid, e.g. straight
```
OnnoEbbens, line 65:
```suggestion
        A single lake may have multiple elements (polygon pieces) in one cell, e.g. straight
```

OnnoEbbens (changes requested):
Thanks Bas, hadn't realised this was happening for lakes. I have just some small textual comments.

Another thing is that I had a bit of a hard time understanding the PR text. It only clicked after I looked into the code. It looks as if it is AI generated. Not necessarily something I oppose but maybe the AI can make it more concise next time :)

OnnoEbbens (changes requested):
Thanks Bas, hadn't realised this was happening for lakes. I have just some small textual comments.

Another thing is that I had a bit of a hard time understanding the PR text. It only clicked after I looked into the code. It looks as if it is AI generated. Not necessarily something I oppose but maybe the AI can make it more concise next time :)

rubencalje (changes requested):
This is a good improvement. Existing models will change, as the bedleak (conductance) at the edge of lakes is now scaled by the part of the cell that is covered by the lake. This is a good thing though, and will maybe help with model convergence as well.

When Onno's comments are addressed, we can merge this.

bdestombe:
Thanks both! I applied Onno's suggestions, merged the latest dev (CI should pass again) and shortened the description.

PR #585, opened 2026-07-22
Add read LGN utilities

dbrakenhoff:
- automatic download is not possible, so `download_lgn` just prints URL for obtaining download link :sweat_smile: 
- get color map and norm for pretty plotting and get_lgn_legend for creating a matplotlib legend
- lgn_to_grid computes area fractions per LGN code
- simplify lgn with remap_lgn
- load_lgn_within_extent to clip lgn tif to model extent

### DIFF
--- a/nlmod/read/lgn.py
+++ b/nlmod/read/lgn.py
@@ -1,0 +1,391 @@
+# %%
+from pathlib import Path
+from typing import Literal
+
+import geopandas as gpd
+import matplotlib.colors as mcolors
+import matplotlib.patches as mpatches
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+from nlmod.dims.grid import modelgrid_from_ds
+from nlmod.epsg28992 import EPSG_28992
+
+# %%
+
+# Rebuild with QGIS:
+# Load LGN.qlr file > Properties > Symbology > Export Color Map to File...
+# Run nlmod.read.lgn._load_clr_file(<path to clr file>) to get this dictionary
+LGN_COLOR_DICT = {
+    1: {"color": "#73df1f", "rgb": (115, 223, 31), "label": "agrarisch gras"},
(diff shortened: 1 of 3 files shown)


nlmod/read/lgn.py
```diff
@@ -0,0 +1,391 @@
+# %%
+from pathlib import Path
```
OnnoEbbens, line 2: I think this is not an nlmod dependency yet. I am happy to add it though

nlmod/read/lgn.py
```diff
@@ -0,0 +1,391 @@
...
+
+    """
+    return [
+        mpatches.Patch(color=info["color"], label=f"{code}: {info['label']}")
+        for code, info in LGN_COLOR_DICT.items()
+        if code in np.unique(lgn)
+    ]
+
+
+def lgn_to_grid(
```
OnnoEbbens, line 326: I would have expected from the name to get a DataArray with the LGN. Maybe call it `lgn_to_gdf`.

OnnoEbbens (commented):
Nice addition. I had just some small comments. I think ruff also has some complaints.

Also interesting to see the `zonal_crosstab` and the `exact_extract` functions. I think it would be nice at some point to have an overview of the functions we use to go from grid data to vector data (the inverse of https://nlmod.readthedocs.io/en/stable/utilities/06_gridding_vector_data.html).

nlmod/read/lgn.py
dbrakenhoff: isnt pathlib part of std lib? Or did you mean another dependency?

nlmod/read/lgn.py
dbrakenhoff: Just discussed the name is okay. This is similar to `nlmod.grid.gdf_to_grid()` which also returns a GeoDataFrame with results per grid cell. When we convert input to grid and return a DataArray the method name should preferably contain `_to_da`?

PR #588, opened 2026-07-22
allow missing layer definitions in combine_layers:

dbrakenhoff:
- makes it easier to get consistent mappings across multiple extents, just use the mapping for largest extent for the smallest extent and automatically discard missing layers.

### DIFF
--- a/docs/utilities/04_modifying_layermodels.ipynb
+++ b/docs/utilities/04_modifying_layermodels.ipynb
@@ -289,10 +289,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "combine_layers = [\n",
-    "    tuple(np.argwhere(regis.layer.str.startswith(\"URz\").data).squeeze().tolist()),\n",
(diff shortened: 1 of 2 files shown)


OnnoEbbens (approved):
Nice, good addition!

One question. I often use the `combine_layer` dictionary to create a new layer model which is mostly a combination of multiple layers but some layers in my new model are just a single layer in my old model. Often I do want to rename these layers to conform to the new layer model names. Currently if a key of the 'combine_layers' dictionary only refers to a single layer it is ignored because nothing has to be combined. This also means the layer is not renamed. For me it would be really nice if the layer is renamed in this case.

This is especially nice in cases where I want to play with combining layers. If I make a change that result in a single layer, the names of my layer model change kind of unexpectedly causing errors further later in my script.

I am not sure how hard it is to implement and not something that should be implemented necessarily in this PR. I approved this PR so you can already merge if you want to.

PR #592, opened 2026-08-27
add ruff lint step

dbrakenhoff:
remove codacy reporter
format ci files


dbrakenhoff:
This is ready for review I think, mostly fixing ruff issues, which I let Vibe do for me mostly. 

The other is a fix to deal with a current issue in flopy (I think) where the `modelgrid_to_ds()` from a grb file is returning incorrect dimensions for the model top for structured grids.

OnnoEbbens (approved):
I decided to trust the Vibe and not check all 67 files :)

PR #598, opened 2026-09-18
Fix crs of buidrain data

rubencalje:
The original data for pipe drainage (buisdrainage) contains a wrong coordinate system, causing the data to be shifted to the northeast by more than 100 m. This PR fixes this issue by adding the argument `fix_crs` to `nlmod.read.nhi.discretize_buisdrainage`. The value of this argument by default is `False`, so existing models that use this method will not change, but the user will see a warning.

This PR also solves some of the `Use "@" matmul instead of "*"` warnings, and a bathemetry-error.

The following code:

```
import nlmod
extent = [115_000, 122_000, 437_000, 442_000]
ds = nlmod.get_ds(extent, delr=25)
pathname = r"buidrain_test\cache"
ds_out = nlmod.read.nhi.discretize_buisdrainage(ds, pathname, fix_crs=False)

f, ax = nlmod.plot.get_map(extent)
da = ds_out['buisdrain_cond']
nlmod.plot.data_array(da.where(da> 0), ds=ds, ax=ax, alpha=0.5)
backgrounds.BrtAchtergrondKaart(ax=ax).plot(alpha=0.5)
```
produces the image below. The data is placed too far northeast, see the offset near the roads and the pixels in the river.
image

When we set `fix_crs` to `True` in the code above, we get the following image, in which you can see that the data now aligns with the background map.

image


nlmod/read/nhi.py
```diff
@@ -114,6 +115,11 @@ def add_buisdrainage(
     depth_method : str, optional
         The method to transform the depth of buisdrainage to the model Dataset. The
         default is "mode".
+    fix_crs : bool, optional
```
OnnoEbbens, line 118: maybe add a reference to this PR. The image in the PR speaks a thousand bookparts.

pyproject.toml
```diff
@@ -84,6 +83,8 @@ rtd = [
     "sphinx_rtd_theme",
     "nbconvert",
     "openpyxl",
+    "numpy<2.5.0", # until there is a new version of netCDF4 (now 1.7.4) this solves "Setting the shape on a NumPy array has been deprecated"
+    "affine<3.0.1", # until there are newer versions of rasterio (now 1.5.1) and rioxarray (now 0.23.0), this solves "Use @ matmul instead of * mul operator for matrix multiplication"
```
OnnoEbbens, line 87: But you did fix the matmul right? Is this <3.0.1 still necessary?

nlmod/read/rws.py
OnnoEbbens, line 1: Not sure what is happening with these changes but I assume it is some fix for RWS bathymetry download?

OnnoEbbens (approved):
Nice work! I just have some minor remarks.

nlmod/read/rws.py
rubencalje: Yes, the FeatureServer is not avialable anymore, but the MapServer is.  Then in Python 3.12, the stuff with the env-variable GDAL_DISABLE_READDIR_ON_OPEN was needed, otherwise it would look for non-existing files and fail.

pyproject.toml
rubencalje: I fixed it in our own code. But it still needs to be fixed in methods within rasterio and rioxarray, that we use.

nlmod/read/nhi.py
rubencalje: good idea, will add a link to the docstring

